PR #115, opened 2018-08-30
Tests for Logutils are finished

1522784:


FabianK1991:
These tests should be adjusted since the log package was refactored

PR #282, opened 2018-10-09
Update README.md, fix installation script for Mac

zkhadikov:
Curl needs "Accept: application/json" in request header, else  call 
curl "https://api.github.com/repos/covexo/devspace/releases/latest" fails with following error message 
{
  "message": "Unsupported 'Accept' header: [#<Sinatra::Request::AcceptEntry:0x00007f8e4e2f8878 @entry=\"application/vnd.heroku+json; version=3\", @type=\"application/vnd.heroku+json\", @params={\"version\"=>\"3\"}, @q=1.0>]. Must accept 'application/json'.",
  "documentation_url": "https://developer.github.com/v3/media"
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ curl --silent "https://api.github.com/repos/covexo/devspace/releases/latest" | s
 
 ### For Mac
 ```bash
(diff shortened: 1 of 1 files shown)


FabianKramm:
Thanks for opening this pull request! You are right, if the standard accept http header is somehow overridden, the installation script fails, hence I will merge this pull request.

PR #437, opened 2018-12-28
Feature/go modules

KaelBaldwin:
/kind feature

None

No

This will change from using dep for dependency management to the built in go modules.

go 1.11.4 is required for this.

I had an issue with the dependency: github.com/otiai10/copy @ commmit: 7e9a647135a1

While testing this on both windows and linux, for some odd reason I kept getting a hash checksum mismatch error for that one, so I reverted one commit to 0046ee23fdbd and then everything worked fine.

I don't think anything really changed between the two commits, but you may want to look more closely at that.

I believe I was able to maintain all the constraints from the Gopkg.toml file.

Benefits of this switch are faster dependency loading time and also this is the official go tool for dependency management now.

Also, using go modules allows you to work outside of your gopath, there may be some things you have to do if you want to continue working within your gopath (check the go module documentation), but I recommend working outside of it instead.


KaelBaldwin:
I understand that some scripts will need to be adjusted to use go 1.11.4 before this can be merged as well.

KaelBaldwin:
The README should probably be updated as well. I'll work on that.

FabianKramm:
LGTM

PR #546, opened 2019-05-21
Test envutil

1522784:
/kind enhancement

It shows a bug in envutil: AddToPath() doesn't work for some reason (On my windows machine, could be different in travis)

no

no

nothing


FabianKramm:
No test currently required, we only wrap vendor calls in this package.

PR #613, opened 2019-07-31
resolves #612

### LINKED ISSUE #612: Allow login without browser instead of waiting forever
When DevSpace tries to open the browser to show the login page, it may occur that the CLI is waiting forever because the user does not have a browser installed (e.g. when using DevSpace in a container or a VM).

gcloud handles this by detecting the failed attempt to open the browser and showing a URL to the user for manual authentication. After user logs in, a auth key is shown and can be copied to the gcloud command which is asking for this key.

Alternative would be aborting with an error message but the usability of this is not great especially when users run `devspace init` and already initialized their project half-way.

**Additional context**  

/kind feature

LukasGentele:
/kind enhancement

resolves #612 => This could be solved even nicer, e.g. gcloud directly generates a key and displays it, so users can directly copy it after logging in. But for now, this is a working solution which only requires to click 2 additional buttons before being able to copy an access key. That should be acceptable for now and is already much better than the current behavior.

Yes. The cloud login asks for an access key now if the user does not have a browser installed or DevSpace fails opening it.

No.

There has been a commit for DevSpace Cloud to allow redirects after logging in, so we can send users directly to /settings/access-keys without having to explain a lot.

### DIFF
--- a/pkg/devspace/cloud/login.go
+++ b/pkg/devspace/cloud/login.go
@@ -3,6 +3,7 @@ package cloud
 import (
 	"context"
 	"fmt"
+	"strings"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


FabianKramm:
LGTM

PR #656, opened 2019-08-30
Improve init and dev

### LINKED ISSUE #655: devspace init: don't override entrypoint for php containers
If you create a new devspace.yaml via `devspace init` the dev configuration shouldn't override the entrypoint

LukasGentele:
/kind enhancement

This PR changes the behavior of init (see below) and adds a the `--interactive / -i` flag to `devspace dev`.

Also resolves #655 

Yes:
- `init` warns when application listens on a system port and asks for forwarding port
- `init` now writes a different config which by default disables the terminal and does not override the entrypoint anymore

No

Due to a cobra issue (https://github.com/spf13/cobra/issues/866), the --interactive flag must be used as `-i/--interactive=value` instead of `-i/--interactive value` (or alternatively without value at all).

### DIFF
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -423,7 +423,7 @@ func (r *reloadError) Error() string {
 
 func (cmd *DevCmd) loadConfig(generatedConfig *generated.Config) *latest.Config {
 	// Get config with adjusted cluster config
-	config, err := configutil.GetContextAjustedConfig(cmd.Namespace, "")
+	config, err := configutil.GetContextAdjustedConfig(cmd.Namespace, "", false)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -142,7 +142,7 @@ func (cmd *DevCmd) Run(cobraCmd *cobra.Command, args []string) {
 	config := cmd.loadConfig(generatedConfig)
 
 	// Signal that we are working on the space if there is any
-	err = cloud.ResumeSpace(config, generatedConfig, true, log.GetInstance())
+	err = cloud.ResumeLatestSpace(config, true, log.GetInstance())
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -57,7 +57,7 @@ type DevCmd struct {
 	Namespace       string
 }
 
-var interactiveDefaultPickerValue = "Open Picker"
+const interactiveDefaultPickerValue = "Open Picker"
 
 // NewDevCmd creates a new devspace dev command
 func NewDevCmd() *cobra.Command {
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -25,6 +25,8 @@ import (
 	"github.com/devspace-cloud/devspace/pkg/devspace/services"
 	"github.com/devspace-cloud/devspace/pkg/util/analytics/cloudanalytics"
 	"github.com/devspace-cloud/devspace/pkg/util/log"
+	"github.com/devspace-cloud/devspace/pkg/util/ptr"
+	"github.com/devspace-cloud/devspace/pkg/util/survey"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
 )
@@ -48,11 +50,14 @@ type DevCmd struct {
 	SwitchContext   bool
 	Portforwarding  bool
 	VerboseSync     bool
+	Interactive     string
 	Selector        string
 	Container       string
 	LabelSelector   string
 	Namespace       string
 }
+
+var interactiveDefaultPickerValue = "Open Picker"
 
 // NewDevCmd creates a new devspace dev command
 func NewDevCmd() *cobra.Command {
@@ -102,6 +107,11 @@ Starts your project in development mode:
 
 	devCmd.Flags().BoolVar(&cmd.SwitchContext, "switch-context", true, "Switch kubectl context to the DevSpace context")
 	devCmd.Flags().BoolVar(&cmd.ExitAfterDeploy, "exit-after-deploy", false, "Exits the command after building the images and deploying the project")
+
+	devCmd.Flags().StringVarP(&cmd.Interactive, "interactive", "i", interactiveDefaultPickerValue, "Enable interactive mode for images (overrides entrypoint with sleep command) and start terminal proxy")
+
+	// Allows to use `devspace dev -i` without providing a value for the flag, see https://github.com/spf13/pflag#setting-no-option-default-values-for-flags
+	devCmd.Flags().Lookup("interactive").NoOptDefVal = interactiveDefaultPickerValue
 
 	return devCmd
 }
@@ -432,6 +442,74 @@ func (cmd *DevCmd) loadConfig(generatedConfig *generated.Config) *latest.Config 
 	err = generated.SaveConfig(generatedConfig)
 	if err != nil {
 		log.Fatalf("Couldn't save generated config: %v", err)
+	}
+
+	// Adjust config for interactive mode
+	if cmd.Interactive != "" {
+		if config.Images == nil {
+			log.Fatal("Your configuration does not contain any images to build for interactive mode. If you simply want to start the terminal instead of streaming the logs, run `devspace dev -t`")
+		}
+		images := *config.Images
+
+		if cmd.Interactive == interactiveDefaultPickerValue {
+			imageNames := make([]string, 0, len(images))
+			for k := range images {
+				imageNames = append(imageNames, k)
+			}
+
+			// If only one image exists, use it, otherwise show image picker
+			if len(imageNames) == 1 {
+				cmd.Interactive = imageNames[0]
+			} else {
+				cmd.Interactive = survey.Question(&survey.QuestionOptions{
+					Question:     "Which image do you want to build using the 'ENTRPOINT [sleep, 999999]' override?\nIf you want to apply this override to multiple images run `devspace dev -i image1,image2,...`",
+					DefaultValue: useDevSpaceCloud,
+					Options:      imageNames,
+				})
+			}
+		}
+
+		// Make sure dev section exists in config
+		if config.Dev == nil {
+			config.Dev = &latest.DevConfig{}
+		}
+
+		// Make sure dev.overrideImages section exists in config
+		if config.Dev.OverrideImages == nil {
+			imageOverrideConfig := []*latest.ImageOverrideConfig{}
+			config.Dev.OverrideImages = &imageOverrideConfig
+		}
+		imageOverrideConfig := *config.Dev.OverrideImages
+
+		// Entrypoint used for interactive mode
+		entrypointOverride := []*string{
+			ptr.String("sleep"),
+			ptr.String("999999"),
+		}
+
+		// Set all entrypoint overrides for specified interactive images
+		interactiveImages := strings.Split(cmd.Interactive, ",")
+		for i := range interactiveImages {
+			imageName := strings.TrimSpace(interactiveImages[i])
+			if _, ok := images[imageName]; !ok {
+				log.Fatalf("Unable to find image '%s' in configuration", imageName)
+			}
+			imageOverrideConfig = append(imageOverrideConfig, &latest.ImageOverrideConfig{
+				Name:       &imageName,
+				Entrypoint: &entrypointOverride,
+			})
+			log.Infof("Interactive mode: override image %s with 'ENTRYPOINT [sleep, 999999]'", imageName)
+		}
+		config.Dev.OverrideImages = &imageOverrideConfig
+
+		// Make sure dev.terminal section exists in config
+		if config.Dev.Terminal == nil {
+			config.Dev.Terminal = &latest.Terminal{}
+		}
+
+		// Set dev.terminal.disabled = false
+		config.Dev.Terminal.Disabled = ptr.Bool(false)
+		log.Info("Interactive mode: enable terminal")
 	}
 
 	return config
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -18,7 +18,6 @@ import (
 	"github.com/devspace-cloud/devspace/pkg/devspace/config/configutil"
 	"github.com/devspace-cloud/devspace/pkg/devspace/config/generated"
 	latest "github.com/devspace-cloud/devspace/pkg/devspace/config/versions/latest"
-	v1 "github.com/devspace-cloud/devspace/pkg/devspace/config/versions/latest"
 	"github.com/devspace-cloud/devspace/pkg/devspace/docker"
 	"github.com/devspace-cloud/devspace/pkg/devspace/kubectl"
 	"github.com/devspace-cloud/devspace/pkg/devspace/registry"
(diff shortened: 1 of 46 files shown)


pkg/devspace/configure/deployment.go
```diff
@@ -76,6 +77,20 @@ func GetDockerfileComponentDeployment(config *latest.Config, generatedConfig *ge
...
 
+		if port < 1024 {
+			log.Warn("Your application listens on a system port [0-1024]. Choose a forwarding-port to access your application via localhost.")
+
+			portString := survey.Question(&survey.QuestionOptions{
+				Question:     "Which forwarding port [1024-49151] do you want to use to access your application?",
+				DefaultValue: strconv.Itoa(port + 8000),
+			})
+
+			port, err = strconv.Atoi(portString)
```
FabianKramm: This will potentially lead to wrong configuration since an application port often cannot chosen at will. Rather we should set the specified port here as the port of the service and the other port as the target port the service will map incoming connections onto

pkg/devspace/configure/deployment.go
```diff
@@ -57,7 +58,7 @@ func GetDockerfileComponentDeployment(config *latest.Config, generatedConfig *ge
 			port = strconv.Itoa(ports[0])
 		} else if len(ports) > 1 {
 			port = survey.Question(&survey.QuestionOptions{
-				Question:     "Which port is the container listening on?",
+				Question:     "Which port is your application listening on?",
```
FabianKramm: This could lead to potential confusion since the function is also used for `devspace add deployment`. The user could think he now has to enter the port for the whole application instead the port for the single container he added is asked

go.mod
```diff
@@ -91,6 +91,7 @@ require (
 	k8s.io/apimachinery v0.0.0
 	k8s.io/client-go v0.0.0
 	k8s.io/helm v2.14.2+incompatible
+	k8s.io/klog v0.3.1
```
FabianKramm: Why are there changes to the go.mod?

cmd/dev.go
```diff
@@ -103,6 +108,11 @@ Starts your project in development mode:
 	devCmd.Flags().BoolVar(&cmd.SwitchContext, "switch-context", true, "Switch kubectl context to the DevSpace context")
 	devCmd.Flags().BoolVar(&cmd.ExitAfterDeploy, "exit-after-deploy", false, "Exits the command after building the images and deploying the project")
 
+	devCmd.Flags().StringVarP(&cmd.Interactive, "interactive", "i", interactiveDefaultPickerValue, "Enable interactive mode for images (overrides entrypoint with sleep command) and start terminal proxy")
```
FabianKramm: I'm not sure if we should name the flag interactive, since the major change is that an image will be rebuild using a different entrypoint. I prefer 'sleep' as an option, but we should talk about this.

cmd/dev.go
```diff
@@ -434,5 +444,73 @@ func (cmd *DevCmd) loadConfig(generatedConfig *generated.Config) *latest.Config
 		log.Fatalf("Couldn't save generated config: %v", err)
 	}
 
+	// Adjust config for interactive mode
+	if cmd.Interactive != "" {
+		if config.Images == nil {
```
FabianKramm: There should be also a check for len(*config.Images) == 0, since it is possible to have an empty map for images

cmd/dev.go
```diff
@@ -434,5 +444,73 @@ func (cmd *DevCmd) loadConfig(generatedConfig *generated.Config) *latest.Config
...
+				imageNames = append(imageNames, k)
+			}
+
+			// If only one image exists, use it, otherwise show image picker
+			if len(imageNames) == 1 {
+				cmd.Interactive = imageNames[0]
+			} else {
+				cmd.Interactive = survey.Question(&survey.QuestionOptions{
+					Question:     "Which image do you want to build using the 'ENTRPOINT [sleep, 999999]' override?\nIf you want to apply this override to multiple images run `devspace dev -i image1,image2,...`",
+					DefaultValue: useDevSpaceCloud,
```
FabianKramm: This seems wrong useDevSpaceCloud is probably not a real image name

cmd/dev.go
LukasGentele: I thought about that as well but `sleep` has some issues:
- It is very confusing because of the sleep mode for spaces. 
- It is too implementation-specific. We are using the sleep command for now but we could also do something like `cat` or `tail -f /dev/null` instead. They do the same but are not related to sleep.
- Additionally, the terminal is also enabled to having a name that expresses more than just build-related terminology makes sense.

The flag should be named after what the user wants to achieve with it. `interactive` expresses that the user has to do something and a terminal session is as interactive as it gets. The only alternative I could think of is `manual` because it says that the user has to manually start the container. That would make sense as well but there is not much of a difference between `manual` and `interactive`. More a matter of preference, I think. I would be ok with both of them.

go.mod
LukasGentele: Not sure. I just ran `go build`. Sometimes the go.mod changes as a result. Maybe the module support in the newest go version does anything there? Removing the `indirect` is not a big deal but I don't know where `klog` comes from because it is not used anywhere in the files I edited. But it is used in `pkg/devspace/kubectl/transport/connection.go`, so it should have been in there beforehand. So, maybe go or the formatter in VS code fixes such things automatically now.

pkg/devspace/configure/deployment.go
LukasGentele: Not sure if application always perceived as the entire thing. I understand the concern more for `devspace add deployment` but honestly, no one uses that command and in `init` the question for container is confusing because there is no container, yet. At most, there is a Dockerfile. And I think asking for ports of containers that do not exist, may be confusing, especially if a user did not even have a Dockerfile beforehand.

pkg/devspace/configure/deployment.go
LukasGentele: That is true. This should be moved to init.addDevConfig and do what it says in the question. Done.

cmd/init.go
```diff
@@ -436,10 +437,29 @@ func (cmd *InitCmd) addDevConfig() {
...
+				if err != nil {
+					log.Fatal("Error parsing port '%s'", portString)
+				}
+			}
 			portMappings := []*latest.PortMapping{}
-			exposedPort := *servicePort.Port
 			portMappings = append(portMappings, &latest.PortMapping{
-				LocalPort: &exposedPort,
+				LocalPort:  &localPort,
+				RemotePort: &remotePort,
```
FabianKramm: This will result in a wrong port mapping since remotePort is 0 if not set and the port mapping localPort -> 0 is invalid

cmd/dev.go
```diff
@@ -48,12 +50,15 @@ type DevCmd struct {
...
 	Portforwarding  bool
 	VerboseSync     bool
+	Interactive     string
 	Selector        string
 	Container       string
 	LabelSelector   string
 	Namespace       string
 }
 
+var interactiveDefaultPickerValue = "Open Picker"
```
FabianKramm: make this a const

LukasGentele: Done

cmd/init.go
LukasGentele: Using pointers instead of int values directly. That should do as expected.

cmd/init.go
```diff
@@ -112,8 +116,20 @@ func (cmd *InitCmd) Run(cobraCmd *cobra.Command, args []string) {
 	// Print DevSpace logo
 	log.PrintLogo()
 
-	// Check if user wants to use devspace cloud
-	cmd.checkIfDevSpaceCloud()
+	cloudRegistryHostname, err := cmd.getCloudRegistryHostname()
```
FabianKramm: Why is this the first question that is asked? When the user does not want to push an image because for example he selectes the use existing image he doesn't need to specify an image registry

cmd/init.go
```diff
@@ -257,146 +274,152 @@ func (cmd *InitCmd) Run(cobraCmd *cobra.Command, args []string) {
...
-	// Get provider configuration
-	providerConfig, err := cloudconfig.ParseProviderConfig()
-	if err != nil {
-		log.Fatalf("Error loading provider config: %v", err)
-	}
+	// Check if user has specified a certain provider
+	var cloudProvider *string
+	if cmd.Provider == "" {
+		// prevents EnsureLoggedIn call in GetProvider
+		registryURL = "dscr.io"
```
FabianKramm: This shouldn't be hardcoded since this information is retrieved from the server. This is also dependent on which provider is used

cmd/init.go
```diff
@@ -257,146 +274,152 @@ func (cmd *InitCmd) Run(cobraCmd *cobra.Command, args []string) {
...
+		registryURL = dockerHubHostname
+		registryLoginHint = fmt.Sprintf(registryLoginHint, "")
+	} else if selectedRegistry == useDevSpaceRegistry {
+		registryURL = cloudRegistryHostname
+		registryLoginHint = fmt.Sprintf(registryLoginHint, " "+cloudRegistryHostname)
+		cmd.useCloud = true
+	} else {
+		registryURL = survey.Question(&survey.QuestionOptions{
+			Question:     "Please enter the registry URL without image name:",
+			DefaultValue: "http://my.registry.tld/username",
```
FabianKramm: Should be without http://.

cmd/init.go
```diff
@@ -257,146 +274,152 @@ func (cmd *InitCmd) Run(cobraCmd *cobra.Command, args []string) {
...
-			cmd.connectCluster()
+	if dockerClient != nil {
+		log.StartWait("Checking registry authentication")
+		authConfig, err := docker.Login(dockerClient, registryURL, "", "", true, false, false)
+		log.StopWait()
+
+		if err != nil || authConfig.Username == "" {
+			log.Fatalf("Registry authentication failed for %s.\n         %s", registryURL, registryLoginHint)
 		}
+	} else if selectedRegistry != useDevSpaceRegistry {
```
FabianKramm: Why check for useDevSpaceRegistry here? The user will get an error on provider login then since dockerClient cannot be created

cmd/init.go
```diff
@@ -257,146 +274,152 @@ func (cmd *InitCmd) Run(cobraCmd *cobra.Command, args []string) {
...
+	registryUsernameHint := " => you are logged in as %s"
+	registryDefaultOption := useDevSpaceRegistry
+	registryLoginHint := "Please login via `docker login%s` and try again."
 
-		// Create generated yaml if cloud
-		generatedConfig, err := generated.LoadConfig()
-		if err != nil {
-			log.Fatal(err)
+	config := configutil.GetConfig()
+	dockerClient, err := docker.NewClient(config, false, log.GetInstance())
```
FabianKramm: This should throw an error because something with docker is messed up. If this fails it does not mean docker is not installed it usually means there are some wrong environment variables set and we should inform the user about this because it will lead to an error later on anyways

cmd/use/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+
+func newNamespaceCmd() *cobra.Command {
+	cmd := &namespaceCmd{}
+
+	useNamespace := &cobra.Command{
+		Use:   "namespace",
+		Short: "Tells DevSpace which namespace to deploy to",
+		Long: `
+#######################################################
+################ devspace use space ###################
```
FabianKramm: devspace use namespace

cmd/use/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+#######################################################
+Set the default namesapce to deploy to
```
FabianKramm: naesapce -> namespace

cmd/use/namespace.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+type namespaceCmd struct{}
+
+func newNamespaceCmd() *cobra.Command {
+	cmd := &namespaceCmd{}
+
+	useNamespace := &cobra.Command{
+		Use:   "namespace",
+		Short: "Tells DevSpace which namespace to deploy to",
```
FabianKramm: I'm not sure if we need this command since in my eyes it makes things more complex for the user and us. The main problem with this command (and `devspace use space`) is that it hides what kube context and namespace is used during deployment. While for `devspace use space` this is partly encouraged since the kube context and namespace are generated by us, a command like this would lead to much confusion since a user doesn't sees the used kube context or namespace in his `devspace.yaml` anymore and has to guess which namespace and kube context he had configured.  

If we decide for such a command, I would prefer to change the devspace.yaml directly like other configure commands are doing which is much more transparent for the user and also makes the code a lot simpler.

pkg/devspace/config/configutil/get.go
```diff
@@ -500,3 +509,55 @@ func GetDefaultNamespace(config *latest.Config) (string, error) {
...
+	generatedConfig, err := generated.LoadConfig()
+	if err != nil {
+		return nil, errors.Errorf("Error loading generated.yaml: %v", err)
+	}
+
+	// load config
+	config, err := GetConfigFromPath(".", generatedConfig.ActiveConfig, true, generatedConfig, log.GetInstance())
+	if err != nil {
+		log.Fatal(err)
+	}
```
FabianKramm: The generated config has to be saved in this function otherwise entered variables will not be saved.

pkg/devspace/kubectl/util.go
```diff
@@ -65,15 +66,22 @@ func EnsureDefaultNamespace(config *latest.Config, client kubernetes.Interface,
 	if defaultNamespace != "default" {
```
FabianKramm: Use the correct kubernetes constant for this

pkg/devspace/kubectl/util.go
```diff
@@ -65,15 +66,22 @@ func EnsureDefaultNamespace(config *latest.Config, client kubernetes.Interface,
...
 			// Create release namespace
 			_, err = client.CoreV1().Namespaces().Create(&v1.Namespace{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: defaultNamespace,
 				},
 			})
+
+			log.Donef("Created namespace %s", defaultNamespace)
 		}
+	} else {
```
FabianKramm: This check isn't working as intended, since in the message the command `devspace use namespace` will change the default namespace but the warning will be printed again because configutil.GetDefaultNamespace will return default again

cmd/init.go
LukasGentele: Done

cmd/init.go
LukasGentele: But would this return an error if Docker is not installed or not?

cmd/use/namespace.go
LukasGentele: Done

cmd/use/namespace.go
LukasGentele: Done

pkg/devspace/config/configutil/get.go
LukasGentele: Makes sense. Was not obvious from the code. Maybe this should generally done after setting all variables?

LukasGentele: Done for now as requested.

pkg/devspace/kubectl/util.go
LukasGentele: Done

cmd/init.go
LukasGentele: Done

LukasGentele:
closing this in favor of #662 662

PR #667, opened 2019-09-08
fix fatal that occurs when running devspace init

LukasGentele:
/kind bug

prevents fatal when running `devspace init` in v4 beta1


FabianKramm:
LGTM

PR #747, opened 2019-10-16
fix: add port forwarding on init

LukasGentele:
/kind bug

Makes sure DevSpace finds the component deployment and is able to add the port forwarding configuration automatically. Due to the fact that we changed from `component` to `helm` deployment when adding the deployment during `devspace init`, the check that looks for the component was always false resulting in not adding the port forwarding config by default. This PR fixes this.


FabianKramm:
Close in favor of #744

PR #752, opened 2019-10-21
Fixed Windows Powershell Command

afmsavage:
/kind documentation

resolves #

Yes, just fixing the front-page README.md to a more universal Powershell syntax.

No

I changed the Windows Powershell install command to use `Invoke-WebRequest` instead of `wget`.  wget is not an alias that Powershell has set by default in newer PSCore versions.  Tested on PSCore 6 and 7.  Older versions of Powershell for Windows, 5.1, do have `wget` aliased though to `Invoke-WebRequest`.


FabianKramm:
@afmsavage good catch, thanks for fixing! LGTM

PR #794, opened 2019-11-11
Add kubectl client auth plugins

mgruener:
/kind enhancement

**What is this pull request for? Which issues does it resolve?**
Adds support for the different kubectl auth-provider plugins (for example OIDC). Without the plugins devspace fails with the following error during deployments (and other operations that require kubernetes access):

```[fatal]  Unable to create new kubectl client: new client: No Auth Provider found for name "XXX"```

This PR fixes this problem, as suggested by https://github.com/kubernetes/kops/pull/5513

**Does this pull request has user-facing changes?** 
No

Yes, should be visible in the diff.

The actual code change is only the added import in [pkg/devspace/kubectl/client.go](pkg/devspace/kubectl/client.go).


mgruener:
@FabianKramm done :slightly_smiling_face:

PR #810, opened 2019-11-15
Refactor registry pkg

1522784:
/kind enhancement

Refactors the package registry by adding a client that is accessed with an interface


FabianKramm:
Close in favor of #811

PR #834, opened 2019-12-11
Update README.md

DanielGlaas:
Change the second command for installation via Linux Bash from 'mv' to 'install'. By this, the executable will get  the ownership of root and not of the user account

/kind documentation

Improving command for installation in Linux Bash.

No

No


FabianKramm:
@DanielGlaas thanks for opening this pull request! Looks good to me, @LukasGentele we should also change this in the documentation

Issue #949, opened 2020-02-19
docker image not being built (Skip building image)


igoooor:
**What happened?** 
I have one project with two images (named **app-php** and **app-nginx**)
While running `devspace dev` I noticed one of my image (**app-php**) not being built when it should have been.
So to be sure, I did the following:
```
devspace cleanup images
rm -rf ./.devspace
```
Then I ran `devspace dev` again, and the same image was skipped again.
This only happened recently, and I have not clue why.
I took that same project, installed it on a fresh computer, installed devspace as well (first time running it), and the same image was skipped again.
If I remove my second image (**app-nginx**) which was originally built, and I run `devspace dev` again, then my first image (**app-php**) is built.
But if I clean everything again, put back the nginx image, and run `devspace dev` again, then the php image is skipped....

if I use the option `-b` something weirder happens, it builds the nginx twice, see full output:
```
$ devspace --no-warn --config=some-path/devspace.yaml dev --exit-after-deploy --build-sequential -b
[info]   Using kube context 'docker-for-desktop'
[info]   Using namespace 'default'
[info]   Building image 'some-registry/nginx:latest' with engine 'docker'
Sending build context to Docker daemon    209MB
Step 1/3 : FROM some-registry/nginx-base:latest
 ---> 09a2ef174c31
Step 2/3 : COPY --chown=www-data:www-data ops/env /home/www-data/ops/env
 ---> Using cache
 ---> c2328c8d4ae9
Step 3/3 : COPY --chown=www-data:www-data docroot/public /home/www-data/docroot/public
 ---> Using cache
 ---> 37baa8df31e0
Successfully built 37baa8df31e0
Successfully tagged some-registry/nginx:latest
[info]   Skip image push for some-registry/nginx
[done] √ Done processing image 'some-registry/nginx'
[warn]   Newly built image 'some-registry/nginx' has the same tag as in the last build (latest), this can lead to problems that the image during deployment is not updated
[info]   Building image 'some-registry/nginx:latest' with engine 'docker'
Sending build context to Docker daemon    209MB
Step 1/3 : FROM some-registry/nginx-base:latest
 ---> 09a2ef174c31
Step 2/3 : COPY --chown=www-data:www-data ops/env /home/www-data/ops/env
 ---> Using cache
 ---> c2328c8d4ae9
Step 3/3 : COPY --chown=www-data:www-data docroot/public /home/www-data/docroot/public
 ---> Using cache
 ---> 37baa8df31e0
[...]

PR #990, opened 2020-03-02
Testing

FlorianKutz:
/kind enhancement

New Tests.
pkg/build 72%
pkg/kubectl 44%

Additionally if the creation of a namespace fails, the user will no longer see the false confirmation "Created namespace: somenamespace"


pkg/devspace/build/create_builder.go
```diff
@@ -19,14 +19,19 @@ import (
 // createBuilder creates a new builder
 func (c *controller) createBuilder(imageConfigName string, imageConf *latest.ImageConfig, imageTag string, options *Options, log log.Logger) (builder.Interface, error) {
 	var err error
-	var builder builder.Interface
+
+	if c.builder != nil {
```
FabianKramm: This cannot stay, this change caused a severe issue in an older version (#949), because 2 or more images are always build with the same builder.

pkg/devspace/kubectl/client.go
```diff
@@ -212,6 +212,8 @@ func NewClientBySelect(allowPrivate bool, switchContext bool, kubeLoader kubecon
 	return nil, errors.New("We should not reach this point")
 }
 
+var second = time.Second
```
FabianKramm: Why introduce this variable?

pkg/devspace/kubectl/logs.go
```diff
@@ -82,11 +82,13 @@ func (client *client) LogMultipleTimeout(imageSelector []string, interrupt chan
 	for idx, pod := range pods {
 	Outer:
 		for _, container := range pod.Spec.Containers {
+			ImageSelectors:
 			for _, imageName := range imageSelector {
 				if imageName == container.Image {
 					reader, err := client.Logs(ctx, pod.Namespace, pod.Name, container.Name, false, tail, true)
 					if err != nil {
 						log.Warnf("Couldn't log %s/%s: %v", pod.Name, container.Name, err)
+						continue ImageSelectors
```
FabianKramm: Why use a named loop instead of just using continue since it is the most inner loop anyways?

pkg/devspace/kubectl/logs.go
```diff
@@ -82,11 +82,13 @@ func (client *client) LogMultipleTimeout(imageSelector []string, interrupt chan
 	for idx, pod := range pods {
 	Outer:
 		for _, container := range pod.Spec.Containers {
+			ImageSelectors:
```
FabianKramm: see below

pkg/devspace/kubectl/util.go
```diff
@@ -315,8 +316,7 @@ func GetPodStatus(pod *k8sv1.Pod) string {
 	if !initializing {
 		hasRunning := false
 
-		for i := len(pod.Status.ContainerStatuses) - 1; i >= 0; i-- {
-			container := pod.Status.ContainerStatuses[i]
+		for _, container := range pod.Status.ContainerStatuses {
```
FabianKramm: This changes the order the statuses are processed, is the order really not important?

FabianKramm: I would argue the order is very important here, because this could eventually lead to another reason being displayed

pkg/devspace/kubectl/client.go
FlorianKutz: This variable is overwritten in tests with one millisecond to shorten its duration.

pkg/devspace/build/create_builder.go
FlorianKutz: Fixed in commit de974eb355301581d1f5d4158678d0a0d3876a83

pkg/devspace/kubectl/logs.go
FlorianKutz: Honestly I don't think there's a right or wrong here, but I've changed it. 84b6491a110082e58e9800edc7649db1b9820dbd

pkg/devspace/kubectl/util.go
FlorianKutz: Of course this is important. My bad. I reverted the change in a25f9fd7c7535f885b87e6e091c63b07b9c2c8e8

PR #1053, opened 2020-04-05
build: prefer docker over kaniko if both are set

LukasGentele:
/kind enhancement

Because kaniko is fallback to docker, it should be 2nd priority if both are defined. Currently it is 1st priority which does not sounds like a fallback. Inverting this is more understandable and also more expected from a user perspective.

This PR also has the huge advantage that we can now configure docker and still be able to configure the fallback as well. So far, a common use case is that users need `snapshotMode: full` because they have `RUN chmod ..` lines in their Dockerfiles. Without this settings, the kaniko fallback is effectively useless because the resulting image will not be executable due to permission errors. With this PR, users can just configure all kaniko options and they will also be used during the fallback.

Users with both, kaniko and docker configured would see a difference but because kaniko always supersedes docker currently, no one should have both configured right now because the docker options would never be used. So, although the preference will change from kaniko to docker, no one will notice this change.

No

### DIFF
--- a/pkg/devspace/build/create_builder.go
+++ b/pkg/devspace/build/create_builder.go
@@ -86,6 +86,18 @@ func (c *controller) createBuilder(imageConfigName string, imageConf *latest.Ima
 }
 
 func convertDockerConfigToKanikoConfig(dockerConfig *latest.ImageConfig) *latest.ImageConfig {
+	kanikoBuildOptions := &latest.KanikoConfig{
+		Cache: ptr.Bool(true),
+	}
+
+	if dockerConfig.Build != nil && dockerConfig.Build.Kaniko != nil {
+		kanikoBuildOptions = dockerConfig.Build.Kaniko
+	} else {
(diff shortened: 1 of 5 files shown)


pkg/devspace/build/create_builder.go
```diff
@@ -86,6 +86,18 @@ func (c *controller) createBuilder(imageConfigName string, imageConf *latest.Ima
...
 
 func convertDockerConfigToKanikoConfig(dockerConfig *latest.ImageConfig) *latest.ImageConfig {
+	kanikoBuildOptions := &latest.KanikoConfig{
+		Cache: ptr.Bool(true),
+	}
+
+	if dockerConfig.Build != nil && dockerConfig.Build.Kaniko != nil {
+		kanikoBuildOptions = dockerConfig.Build.Kaniko
+	} else {
+		if dockerConfig.Build != nil && dockerConfig.Build.Docker != nil && dockerConfig.Build.Docker.Options != nil {
```
FabianKramm, line 96: else if instead of else & if

PR #1057, opened 2020-04-07
enhancements for init command and restart helper

LukasGentele:
/kind enhancement

- uses CMD for restart helper if ENTRYPOINT not found/defined because that works for most images (e.g. all language specific images) and using CMD instead of ENTRYPOINT is recommended by Docker because ENTRYPOINT is supposed to just be a shell according to Docker best practices. if that is the best practice, we should just warn about using CMD without knowing ENTRYPOINT but not fail because we can assume that most images have a simple shell in ENTRYPOINT
- init now sets restartContainer, preferSyncOverRebuild, injectRestartHelper, production profile and build.docker.options.target (if multi-stage dockerfile)
- improves restart helper script to retry when immediate start script execution fails (important for many web servers who still have open connections that need a bit more time to terminate after KILL)
- fixes dockerfile target split regex

### DIFF
--- a/pkg/devspace/build/builder/helper/util.go
+++ b/pkg/devspace/build/builder/helper/util.go
@@ -117,10 +117,9 @@ func RewriteDockerfile(dockerfile string, entrypoint []string, cmd []string, tar
 		if len(entrypoint) == 0 {
 			if len(oldEntrypoint) == 0 {
 				if len(oldCmd) == 0 {
-					return "", errors.Errorf("Cannot inject restart helper into Dockerfile because neither ENTRYPOINT nor CMD was found.\n\nHow to fix this:\n- Option A: Define an ENTRYPOINT (or CMD) in your Dockerfile\n- Option B: Set `images.*.entrypoint` option in your devspace.yaml")
-				} else {
-					log.Warn("Using CMD statement for injecting restart helper because ENTRYPOINT is missing in Dockerfile and `images.*.entrypoint` is also not configured")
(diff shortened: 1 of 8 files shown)


pkg/devspace/build/builder/helper/util.go
```diff
@@ -111,7 +116,11 @@ func RewriteDockerfile(dockerfile string, entrypoint []string, cmd []string, tar
 
 		if len(entrypoint) == 0 {
 			if len(oldEntrypoint) == 0 {
-				return "", errors.Errorf("cannot inject restart helper into dockerfile because no ENTRYPOINT was found, please make sure your dockerfile has an ENTRYPOINT defined or define it in the devspace.yaml in 'images.*.entrypoint'")
+				if len(oldCmd) == 0 {
+					return "", errors.Errorf("Cannot inject restart helper into Dockerfile because neither ENTRYPOINT nor CMD was found.\n\nHow to fix this:\n- Option A: Define an ENTRYPOINT (or CMD) in your Dockerfile\n- Option B: Set `images.*.entrypoint` option in your devspace.yaml")
```
FabianKramm, line 120: We should use lowecase error strings as this is a golang best practice: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/devspace/build/builder/helper/util.go
```diff
@@ -111,7 +116,11 @@ func RewriteDockerfile(dockerfile string, entrypoint []string, cmd []string, tar
...
+				} else {
```
FabianKramm, line 121: this else is unnecessary since the function returns if oldCmd length is 0

PR #1194, opened 2020-10-12
feat: use helm cli for v3

FlorianKutz:


FabianKramm:
close in favor of #1196

PR #1217, opened 2020-11-17
Fix patches.mdx

harryttd:
/kind documentation

Documentation edit. In memory profile config did not reflect the correct changes that were applied by the profile patch.


FabianKramm:
@harryttd thanks for opening this pull request! Looks good to me, good catch!

PR #1293, opened 2021-01-12
Restructure e2e tests

FlorianKutz:
It completely restructures the end-to-end test.

Here is a full list of the current tests:

Build command:
  Kaniko: This test uses kaniko to build the image

Deploy command:
  Component Chart: Helm is used to deploy a component chart as well as two local dependencies
  Helm v3 with kaniko (Skipped!): This test is supposed to build with kaniko and deploy with helm v3, but Kaniko needs to push the image which isn't possible in the CI pipeline.
  Helm v2 (Skipped!): This test is skipped, because the cli of Helm v2 doesn't work reliably.
  Kube yamls: In this test kubectl is used to deploy the resources.
  Kustomize: This test uses kustomize to deploy.

Dev command:
  custom-builder: This test tries out the command with a custom builder
  default: This test tries out a lot of features of the command: Sync, Port Forwardig, the UI server and automatic pod restarts. However, the port forwarding doesn't work in a ci pipeline, therefore this part is skipped.
  dependencies: This test tries the dev command with some dependencies configured in devspace.yaml.
  hooks: This test uses the dev command with configured hooks. It's skipped on Windows.

Enter command:
  default: This tests enters a simple echo-command into the terminal of a container

Init command:
  Create Dockerfile: This tests uses the command to generate a Dockerfile.
  Everything already created: This test uses the command in a folder where all config files are already there.

Print command:
  Default: This test tests the basic functionality including the vars-flag

Render command:
  Helm v2(Skipped!): This test is skipped, because the cli of Helm v2 doesn't work reliably.
  Helm v3: This test uses the command with a helm deployment in the devspace.yaml
  Kubectl: This test uses the command with a kube deployment in the devspace.yaml


FabianKramm:
Closed because not needed anymore

PR #1310, opened 2021-02-02
feat: add flag ui-port

FlorianKutz:
/kind feature

it adds a new flag

yes, the user can now use a new flag when running devspace dev

no

### DIFF
--- a/cmd/dev.go
+++ b/cmd/dev.go
@@ -1,14 +1,15 @@
 package cmd
 
 import (
-	"github.com/loft-sh/devspace/pkg/devspace/plugin"
-	"github.com/loft-sh/devspace/pkg/devspace/server"
-	"github.com/loft-sh/devspace/pkg/devspace/services"
-	"github.com/loft-sh/devspace/pkg/devspace/upgrade"
 	"os"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/loft-sh/devspace/pkg/devspace/plugin"
+	"github.com/loft-sh/devspace/pkg/devspace/server"
+	"github.com/loft-sh/devspace/pkg/devspace/services"
+	"github.com/loft-sh/devspace/pkg/devspace/upgrade"
 
 	"github.com/loft-sh/devspace/cmd/flags"
 	"github.com/loft-sh/devspace/pkg/devspace/analyze"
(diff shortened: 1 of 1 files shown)


cmd/dev.go
```diff
@@ -399,13 +402,19 @@ func (cmd *DevCmd) startServices(f factory.Factory, config *latest.Config, gener
 	}
 
 	// Open UI if configured
-	if cmd.UI {
+	if cmd.UI || cmd.UIPort != 0 {
```
FabianKramm, line 405: We should stay with only `cmd.UI` check here

cmd/dev.go
```diff
@@ -399,13 +402,19 @@ func (cmd *DevCmd) startServices(f factory.Factory, config *latest.Config, gener
...
 		cmd.UI = false
 		logger.StartWait("Starting the ui server...")
 		defer logger.StopWait()
 
+		var port *int
+		if cmd.UIPort != 0 {
+			port = ptr.Int(cmd.UIPort)
+		}
+		cmd.UIPort = 0
```
FabianKramm, line 414: why is that necessary?

cmd/dev.go
```diff
@@ -399,13 +402,19 @@ func (cmd *DevCmd) startServices(f factory.Factory, config *latest.Config, gener
...
 	// Open UI if configured
-	if cmd.UI {
+	if cmd.UI || cmd.UIPort != 0 {
 		cmd.UI = false
 		logger.StartWait("Starting the ui server...")
 		defer logger.StopWait()
 
+		var port *int
+		if cmd.UIPort != 0 {
+			port = ptr.Int(cmd.UIPort)
```
FabianKramm, line 412: why not just use &cmd.UIPort?

FabianKramm (changes requested):

cmd/dev.go
FlorianKutz: Because otherwise whenever cmd.UIPort == 0 the parameter forcePort isn't nil. It would be a pointer to 0, which forces the port 0 to be used.

cmd/dev.go
FlorianKutz: It was to prevent the server from starting twice. Because of another change, it's not needed anymore.

PR #1321, opened 2021-02-18
vars encryption, logging improvements & survey dependency update

FabianKramm:
### Changes
- devspace now encrypts the entered variables in the `.devspace/generated.yaml`
- updated the internal survey library (https://github.com/AlecAivazis/survey) to v2.2.7
- fixed an issue where the type of variables was not aligned with the default value of that variable
- devspace will now fail during kaniko building if it detects an issue during building pod startup
- devspace will now print a warning during port-forwarding, sync or terminal forwarding if the pod has issues to start
- `devspace init` now allows you to use a registry even though you are not authenticated to it
- fixed an issue where devspace would fail if `--save-vars` was specified and the target namespace was not found


hack/build-all.bash
```diff
@@ -29,7 +29,7 @@ if [[ "$(pwd)" != "${DEVSPACE_ROOT}" ]]; then
 fi
 
 GO_BUILD_CMD="go build -a"
-GO_BUILD_LDFLAGS="-s -w -X main.commitHash=${COMMIT_HASH} -X main.buildDate=${DATE} -X main.version=${VERSION}"
+GO_BUILD_LDFLAGS="-s -w -X main.commitHash=${COMMIT_HASH} -X main.buildDate=${DATE} -X main.version=${VERSION} -X github.com/loft-sh/devspace/pkg/devspace/config/generated.EncryptionKey=$ENCRYPTION_KEY"
```
chenrui333, line 32: why does it need encrypt the key?

relates to https://github.com/Homebrew/homebrew-core/pull/134786

PR #1362, opened 2021-03-25
fix: allow setting base64 encoded variable values

LukasGentele:
/kind bug

Base64 encoded strings often end with `=` and that leads to a problem with this if condition being true and resulting in a fatal.


FabianKramm (approved):
LGTM

PR #1423, opened 2021-04-23
Add option to display helm output during install/upgrade

kuuji:
/kind enhancement
/kind feature

resolves https://github.com/loft-sh/devspace/issues/1419
This adds an option in the helm config to enable output. When set to true this allows to show the output from helm install/upgrade during a deploy.

Since this is technically backward compatible I added it to the current schema but I'm not sure if this actually warrants a new version.

I'm also aware this is missing the documentation part. I wanted to get feedback on the implementation/interface first.

Change should be backward compatible. The field is omitted by default and the behaviour is the same as currently if not provided.
This will require addition to the documentation.

No

Example of what the yaml and the output looks like

```
- name: my-api
  helm:
    chart:
      name: ../helm-charts/charts/my-chart
    output: true
    values:
    valuesFiles:
    - ./values.yaml
```

```
[info]   Execute 'helm upgrade my-api ../helm-charts/charts/my-chart --namespace default --values /var/folders/d4/6jnt6_p51mxgcnhjm9b6fzm80000gp/T/156734643
 --install --kube-context context'
[info]   Helm Output 'Release "my-api" has been upgraded. Happy Helming!
NAME: my-api
LAST DEPLOYED: Fri Apr 23 13:21:23 2021
NAMESPACE: default
STATUS: deployed
REVISION: 2
TEST SUITE: None
NOTES:
Thank you for installing my-chart.

Your release is named my-api.

To learn more about the release, try:

  $ helm status my-api
  $ helm get all my-api
```

### DIFF
--- a/pkg/devspace/config/versions/latest/schema.go
+++ b/pkg/devspace/config/versions/latest/schema.go
@@ -594,6 +594,7 @@ type HelmConfig struct {
 	ValuesFiles      []string                    `yaml:"valuesFiles,omitempty" json:"valuesFiles,omitempty"`
 	ReplaceImageTags *bool                       `yaml:"replaceImageTags,omitempty" json:"replaceImageTags,omitempty"`
 	Wait             bool                        `yaml:"wait,omitempty" json:"wait,omitempty"`
+	Output           bool                        `yaml:"output,omitempty" json:"output,omitempty"`
 	Timeout          *int64                      `yaml:"timeout,omitempty" json:"timeout,omitempty"`
 	Force            bool                        `yaml:"force,omitempty" json:"force,omitempty"`
 	Atomic           bool                        `yaml:"atomic,omitempty" json:"atomic,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/devspace/config/versions/latest/schema.go
```diff
@@ -594,6 +594,7 @@ type HelmConfig struct {
 	ValuesFiles      []string                    `yaml:"valuesFiles,omitempty" json:"valuesFiles,omitempty"`
 	ReplaceImageTags *bool                       `yaml:"replaceImageTags,omitempty" json:"replaceImageTags,omitempty"`
 	Wait             bool                        `yaml:"wait,omitempty" json:"wait,omitempty"`
+	Output           bool                        `yaml:"output,omitempty" json:"output,omitempty"`
```
FabianKramm, line 597: Please rename the option to `DisplayOutput` and `displayOutput`

pkg/devspace/helm/v3/client.go
```diff
@@ -126,7 +127,12 @@ func (c *client) InstallChart(releaseName string, releaseNamespace string, value
 	}
 
 	args = append(args, helmConfig.UpgradeArgs...)
-	_, err = c.genericHelm.Exec(args, helmConfig)
+	output, err := c.genericHelm.Exec(args, helmConfig)
+
+	if helmConfig.Output {
+		c.log.Infof("Helm Output '%s'", output)
```
FabianKramm, line 133: Since there is an output of `[info]   Execute 'helm upgrade ...` just before, I don't think we need a descriptive message here anymore as it should be clear the output corresponds to the helm command. So you could just do:

```
if helmConig.DisplayOutput {
  c.log.Write(output)
}
```

FabianKramm (changes requested):
@kuuji thanks a lot for the PR! I made some comments, if you could change this, we can merge it.

FabianKramm:
@kuuji thanks! LGTM now!

PR #1453, opened 2021-05-31
feat: add new DEVSPACE_GIT_BRANCH_SLUG predefined variable

KeksBeskvitovich:
/kind enhancement
Added DEVSPACE_GIT_BRANCH_SLUG predefined variable. Branch name, where all non alphanumeric symbols replace with  "-". Usful for naming deployment, and helm release, so we can deploy projects with branch-based naming.

no

no


FabianKramm:
@KeksBeskvitovich thanks for creating the pull request, we really appreciate that you want to contribute to DevSpace! 

While we think this might be a useful contribution now, our problem with `DEVSPACE_GIT_BRANCH_SLUG` is that it is essentially just a transformed `DEVSPACE_GIT_BRANCH`, which we would like to solve in future differently (for example with some more advanced templating functionality that allows you to transform variable values with functions). 

If you really need this variable now, you can also write a DevSpace plugin that adds this predefined variable to DevSpace. Please take a look at the [DevSpace docs](https://devspace.sh/cli/docs/guides/plugins#vars) for more information.

vaibhav7797 (approved):
LGTM

PR #1471, opened 2021-06-11
feat: new dependencies[*].forwardVars option

FabianKramm:
### Changes
- New `dependencies[*].overwriteVars` option to disable variable overwriting in dependencies (#1466)
- DevSpace will now automatically reset the parent deployment, statefulset or replicaset during `devspace dev` if a replaced pod cannot be found, but the parent was scaled down already


LukasGentele:
Can we call this passVars, reuseVars, copyVars, populateVars, or something? Forward sounds more like network traffic, port forwarding, email forwarding etc. IMO

FabianKramm:
@LukasGentele yes thats good point, as discussed, we rename it to `overwriteVars`

PR #1485, opened 2021-07-01
Compare ImageName on replacePods

nVitius:
/kind bug

No issue created

no

no

When using multiple entries inside of `dev.replacePods`, this function would fail incorrectly because the comparison was being done on the wrong fields.


FabianKramm:
@nVitius thanks a lot for the PR, yes you are right! Thanks, looks good to me!

PR #1579, opened 2021-08-20
Refer to a container when logging syncing messages

bbuchalter:


pkg/devspace/services/synccontroller/controller.go
```diff
@@ -209,10 +209,15 @@ func (c *controller) startSync(options *Options, onInitUploadDone chan struct{},
 		containerPath = syncConfig.ContainerPath
 	}
 
-	log.Donef("Sync started on %s <-> %s (Pod: %s/%s)", syncClient.LocalPath, containerPath, container.Pod.Namespace, container.Pod.Name)
+	log.Donef("Sync started on %s <-> %s (Container: %s)", syncClient.LocalPath, containerPath, c.containerName())
 	return syncClient, nil
 }
 
+function (c *controller) containerName() string {
```
LukasGentele, line 216: Should be func, not function

Also: Do we really need the namespace here? I think it's mostly useless because most users are operating in 1 namespace and even when using multiple, it is very unlikely that the pod name is not unique. Thoughts? @FabianKramm ?

LukasGentele (changes requested):

Issue #1577, opened 2021-08-19
Name attributes on dev.sync and dev.port configurations


noizwaves:
**Is your feature request related to a problem?**  

Yes. Currently, when using profile patch paths to specify a `dev.sync` or `dev.ports` item to remove, only fragile array indexes can be used. Because there is no "root level" property on these items, robust property selectors cannot be used.

Example (with array indexes):
```
...
dev:
  ports:
  - labelSelector:
      service: web
    forward:
      - port: 3000
        remotePort: 3000
  sync:
    - labelSelector:
        service: web
      excludePaths: []
profiles:
  - name: disable-web
    patches:
      - op: remove
        path: dev.ports[0]
      - op: remove
        path: dev.ports[0]
```

**Which solution do you suggest?**
Adding a new property, like `name` to the `dev.sync` and `dev.ports` configuration. This would allow the above example to be expressed in a robust way:
```
...
dev:
  ports:
  - name: web
    labelSelector:
      service: web
    forward:
      - port: 3000
        remotePort: 3000
  sync:
    - name: web
      labelSelector:
        service: web
      excludePaths: []
profiles:
  - name: disable-web
    patches:
      - op: remove
        path: dev.ports.name=web
      - op: remove
        path: dev.ports.name=web

```

**Which alternative solutions exist?**  
N/A

**Additional context**  
We are using profiles more heavily to control dynamic configuration in our cloud development environments. The ability to safely and reliably combine multiple profiles will be important for us.

FabianKramm:
@noizwaves thanks for creating this issue! Yes that makes sense, this would also help with hooks (like after initial sync) where names can be used to select the specific sync or port forwarding configuration

FabianKramm:
works with v5.16.0

PR #1581, opened 2021-08-23
feat(config): add name to dev.sync & dev.ports

lizardruss:
## Changes

- New option `name` for `dev.sync` and `dev.ports`. This can be used as a steady identifier when using profile patches. (#1577)


FabianKramm:
@lizardruss thanks a lot! Looks good to me!

FabianKramm (approved):

Issue #1512, opened 2021-07-21
Sync stops after initial sync


kickthemooon:
**What happened?**
`devspace sync --config devspace.yaml` command exists after initial sync

```
[info]   Using namespace 'my-namespace'
[info]   Using kube context 'my-context'
[info]   Waiting for pods...
[info]   Starting sync...
[info]   Start syncing
[done] √ Sync started on /local/path <-> /remote/path (Pod: my-pod-name)
[info]   Waiting for initial sync to complete
[info]   Helper - Use inotify as watching method in container
[info]   Downstream - Initial sync completed
[info]   Upstream - Upload 11244 create change(s) (Uncompressed ~38456.75 KB)
[info]   Upstream - Successfully processed 11244 change(s)
[info]   Upstream - Initial sync completed
```

**What did you expect to happen instead?**
expected it to keep running and keep syncing

**How can we reproduce the bug?** (as minimally and precisely as possible)
devspace.yaml

```
version: v1beta10
dev:
  sync:
    - labelSelector:
        app: myapp
      containerName: my-container
      namespace: my-namespace
      containerPath: /app
      uploadExcludePaths:
        - "have/44/paths/under/upload/exclude"
      downloadExcludePaths:
        - "have/75/paths/under/download/exclude"
```

**Local Environment:**

* DevSpace Version: 5.14.4
* Operating System: mac

/kind bug

kickthemooon:
use `devspace dev --config devspace.yaml`

atainter:
Hi, can we reopen this issue? I'm having the same problem on version 5.14.4 for `devspace sync` and `devspace dev`.

When using `devspace dev`, my config looks like this:

```yaml
version: v1beta10

dev:
  sync:
  - labelSelector:
      app.kubernetes.io/name: ${K8S_LABEL_NAME}
      app.kubernetes.io/component: appserver
    containerName: django-uwsgi
    excludePaths:
    - .git/
    - node_modules/
```

This used to work before upgrading versions.

FabianKramm:
@atainter could you post the complete output of `devspace dev`? Does it work if you use [polling](https://devspace.sh/cli/docs/configuration/development/file-synchronization#polling) instead?

FabianKramm:
@DanArlowski does it just stop after the `Initial sync completed` message or how does the output look exactly?

vecerek:
[...]

PR #1618, opened 2021-09-01
fix: devspace sync --config devspace.yaml does not watch

lizardruss:
### Changes
- Fix an issue where `devspace sync` would exit immediately after initial sync and hooks


lizardruss:
I believe this also addresses #1512

Issue #1562, opened 2021-08-10
use .gitignore as the exclude sync files


DanArlowski:
**Is your feature request related to a problem?**
As we use many files and folders in the repo , and every use may have a bit different folder configs, some folders named differently, etc , the gitignore is the most updated file to make sure nothing gets pushed to the repo

**Which solution do you suggest?**
Allow the abillity to state devspace config to use .gitignore as the exclusion list

**Which alternative solutions exist?**
Copy paste by hand, managing 2 different list and keeping them in sync

**Additional context**
May be great feature for complex git repos.

/kind feature

FabianKramm:
@DanArlowski thanks for creating this issue! Yes, I see your point, but I believe this does not always work, however we might want to add an option to load the exclude paths from file, where you could just specify the `.gitignore` file

DanArlowski:
I agree that adding a file may work great,
Then we can add any file and .gitignore as one of them. 

This is a feature I will gladly appreciate.

PR #1619, opened 2021-09-01
feat: use files for excluding paths from synchronization

lizardruss:
### Changes
- Add `excludeFile`, `downloadExcludeFile`, and `uploadExclude` options for loading exclude paths from files using `.gitignore` syntax (#1562)

### DIFF
--- a/pkg/devspace/services/synccontroller/controller.go
+++ b/pkg/devspace/services/synccontroller/controller.go
@@ -4,6 +4,15 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/docker/docker/builder/dockerignore"
 	"github.com/loft-sh/devspace/pkg/devspace/config"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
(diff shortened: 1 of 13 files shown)


pkg/devspace/services/synccontroller/controller.go
```diff
@@ -295,14 +298,38 @@ func (c *controller) initClient(pod *v1.Pod, container string, syncConfig *lates
...
 	if len(syncConfig.UploadExcludePaths) > 0 {
 		options.UploadExcludePaths = syncConfig.UploadExcludePaths
 	}
 
+	if syncConfig.UploadExcludeFile != "" {
+		paths, err := parseExcludeFile(filepath.Join(syncConfig.LocalSubPath, syncConfig.UploadExcludeFile))
+		if err != nil {
+			return nil, errors.Wrap(err, "parse upload exclude file")
+		}
+		options.UploadExcludePaths = append(options.DownloadExcludePaths, paths...)
```
FabianKramm, line 330: Should be options.UploadExcludePaths. Also what happens if `options.UploadExcludePaths` is nil?

FabianKramm (changes requested):
@lizardruss thanks for the PR! Please change the parts I commented on

FabianKramm (approved):
@lizardruss thanks a lot! LGTM!

PR #1638, opened 2021-09-09
Component Chart & UI commands namespace and context

FabianKramm:
### Changes
- Fixed an issue where `pathType` was not automatically set for newer ingress versions
- Fixed an issue where namespace and context were wrong in UI commands
- New flag `--image-selector` for commands `devspace enter`, `devspace logs` & `devspace attach`
- You can now use `image(default)` and `tag(default)` in hook commands and args:
```yaml
images:
  test:
    image: test/test
hooks:
  - command: |-
      ./custom-script image(test):tag(test) # -> Transformed to ./custom-script test/test:##### 
    when:
      after:
        images: all
```
- New feature config expressions `${{ any valid bash expression }}` that are evaluated after variables are loaded and can be used to generate certain fields of the devspace yaml dynamically:

Only build if user uses `devspace deploy`
```yaml
images:
  default:
    build: |-
      ${{ 
        if [ "$1" == "deploy" ]; then 
          echo {disabled: false}
        else
          echo {disabled: true}
        fi
      }}
```
Use devspace var to decide if to open terminal
```yaml
vars:
- name: OPEN_TERMINAL
  options:
  - "true"
  - "false"
dev:
  terminal: ${{ [ "${OPEN_TERMINAL}" == "true" ] && echo "{}" || echo "null" }}
```
Use environment variable to decide if to open terminal:
```yaml
dev:
  terminal: ${{ [ "$OPEN_TERMINAL" == "true" ] && echo "{}" || echo "null" }}
```

### DIFF
--- a/cmd/attach.go
+++ b/cmd/attach.go
@@ -17,6 +17,7 @@ type AttachCmd struct {
 	*flags.GlobalFlags
 
 	LabelSelector string
+	ImageSelector string
 	Image         string
 	Container     string
 	Pod           string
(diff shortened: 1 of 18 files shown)


madhavajay:
It would be great if `hooks` could use the same $ENV variables available to the rest of the file.

PR #1639, opened 2021-09-09
feat: add skip registry option to init

pratikjagrut:
/kind feature

Yes

No

### DIFF
--- a/pkg/devspace/configure/image.go
+++ b/pkg/devspace/configure/image.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
-	v1 "github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
 	"github.com/loft-sh/devspace/pkg/devspace/docker"
 	"github.com/loft-sh/devspace/pkg/devspace/generator"
 	"github.com/loft-sh/devspace/pkg/devspace/pullsecrets"
@@ -20,7 +19,6 @@ import (
 
 const dockerHubHostname = "hub.docker.com"
 const githubContainerRegistry = "ghcr.io"
-const noRegistryImage = "devspace"
 
 // addImage adds an image to the provided config
 func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, dockerfileGenerator *generator.DockerfileGenerator) error {
@@ -28,7 +26,8 @@ func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, doc
 		useDockerHub          = "Use " + dockerHubHostname
 		useGithubRegistry     = "Use GitHub image registry"
 		useOtherRegistry      = "Use other registry"
-		registryDefaultOption = useDockerHub
+		skipRegistry          = "Skip Registry"
+		registryDefaultOption = skipRegistry
 		registryUsernameHint  = " => you are logged in as %s"
 		providedDockerfile    = "Based on this existing Dockerfile: " + dockerfile
 		differentDockerfile   = "Based on a different existing Dockerfile (e.g. ./backend/Dockerfile)"
@@ -41,7 +40,7 @@ func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, doc
 	imageConfig := &latest.ImageConfig{
 		Image:      strings.ToLower(image),
 		Dockerfile: dockerfile,
-		Build: &v1.BuildConfig{
+		Build: &latest.BuildConfig{
 			Disabled: true,
 		},
 	}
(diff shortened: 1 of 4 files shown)


pkg/devspace/configure/image.go
```diff
@@ -136,65 +137,73 @@ func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, doc
...
 			Options:      registryOptions,
 		})
 		if err != nil {
 			return err
 		}
 
-		registryHostname := ""
+		if selectedRegistry != skipRegistry {
+			registryHostname := ""
+			// if selectedRegistry == useDockerHub do nothing.
```
FabianKramm, line 171: This seems not correct to me as now the logic would ask for a registry hostname if you select the docker hub option.

pkg/devspace/configure/image.go
```diff
@@ -136,65 +137,73 @@ func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, doc
...
-			imageConfig.Image = registryUsername + "/" + imageName
-		} else {
-			projectPath := registryUsername + "/project"
-			if regexp.MustCompile("^(.+\\.)?gcr.io$").Match([]byte(registryHostname)) {
-				project, err := exec.Command("gcloud", "config", "get-value", "project").Output()
+			if selectedRegistry == useDockerHub {
+				imageConfig.Image = registryUsername + "/" + imageName
+			} else {
+				projectPath := registryUsername + "/project"
+				if regexp.MustCompile(`^(.+\\.)?gcr.io$`).Match([]byte(registryHostname)) {
```
FabianKramm, line 197: Is this correct? As `"\\"` != `` `\\` ``

pkg/devspace/configure/image.go
```diff
@@ -136,65 +137,73 @@ func (m *manager) AddImage(imageName, image, dockerfile, contextPath string, doc
...
 			useDockerHub = useDockerHub + fmt.Sprintf(registryUsernameHint, authConfig.Username)
 			registryDefaultOption = useDockerHub
+			registryOptions = []string{useDockerHub, useGithubRegistry, useOtherRegistry, skipRegistry}
+			isLoggedIntoRegistry = true
 		}
 
-		authConfig, err = dockerClient.GetAuthConfig(githubContainerRegistry, true)
-		if err == nil && authConfig.Username != "" {
-			useGithubRegistry = useGithubRegistry + fmt.Sprintf(registryUsernameHint, authConfig.Username)
+		if !isLoggedIntoRegistry {
```
FabianKramm, line 150: While this works, it would not show the github user if you are logged in dockerhub and github at the same time

pkg/devspace/configure/image.go
```diff
@@ -20,15 +19,17 @@ import (
 
 const dockerHubHostname = "hub.docker.com"
 const githubContainerRegistry = "ghcr.io"
-const noRegistryImage = "devspace"
+
+// const noRegistryImage = "devspace"
```
FabianKramm, line 23: Is this needed?

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Please see the comments

pkg/devspace/configure/image.go
pratikjagrut: We don't need to escape backslash when we use raw string.
It is recommended to use a raw string in regex. https://staticcheck.io/docs/checks#S1007

pratikjagrut: Removed the extra \.

pkg/devspace/configure/image.go
pratikjagrut: Oh right.

pkg/devspace/configure/image.go
pratikjagrut: So here we want to see the logged-in username and also if a user is logged into docker and Github at the same time we want to make docker as a default value, correct?

pkg/devspace/configure/image.go
pratikjagrut: It is not used in this file and it's not exported as well so I just commented it. Maybe it's good if I remove it.

FabianKramm: @pratikjagrut yes please remove it

pkg/devspace/configure/image.go
FabianKramm: @pratikjagrut yes exactly, there are essentially 4 cases here:

1. Case: User is logged into Dockerhub & Github: Dockerhub Option default and show usernames of Dockerhub and github
2. Case: User is logged into Dockerhub: Dockerhub option default and show username of Dockerhub
3. Case: User is logged into Github: Github option default and show username of Github
4. Case: User is not logged into Dockerhub or Github: Skip registry selected

pkg/devspace/configure/image.go
pratikjagrut: done.

pratikjagrut:
Force pushed for squashing and rebasing the PR.

FabianKramm (approved):
@pratikjagrut looks good to me, thanks!

PR #1647, opened 2021-09-15
feat: allow wildcard and filter paths for patches

lizardruss:
### Changes
- Replaces [yaml-patch](github.com/krishicks/yaml-patch) with [yaml-jsonpath](https://github.com/vmware-labs/yaml-jsonpath)
- Implement `add`, `remove`, and `replace` patch operations using `yaml-jsonpath` as path selectors
- Translates previous `yaml-patch` path syntax into `yaml-jsonpath` filter syntax
- Added unit tests to help ensure backwards compatibility
- Added e2e tests for select edge cases


pkg/devspace/config/loader/patch/operation.go
```diff
@@ -0,0 +1,192 @@
...
+const (
+	opAdd     Op = "add"
+	opRemove  Op = "remove"
+	opReplace Op = "replace"
+)
+
+type Operation struct {
+	Op    Op         `yaml:"op,omitempty"`
+	Path  OpPath     `yaml:"path,omitempty"`
+	From  OpPath     `yaml:"from,omitempty"`
```
FabianKramm, line 23: Is from still supported? If not, we should remove that

pkg/devspace/config/loader/patch/patch.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Patch []Operation
+
+// Apply returns a YAML document that has been mutated per the patch
```
FabianKramm, line 11: per patch instead of per the patch

FabianKramm (changes requested):
@lizardruss thanks a lot for the PR! Looks great, there are just 2 smaller things that should get changed

FabianKramm:
@LukasGentele please check docs

docs/pages/configuration/profiles/patches.mdx
```diff
@@ -115,13 +118,19 @@ profiles:
...
+    value: john/stagingbackend
   - op: remove
     path: deployments.name=backend.helm.values.containers[1]
 - name: production
   patches:
   - op: replace
     path: images.backend.image
     value: john/prodbackend
+  - op: replace
+    path: deployments.name=backend.helm.values.containers[0].image
```
LukasGentele, line 132: Should not use index 0  since we just recommend using selectors just above it. Same for the other one one in this example.

docs/pages/configuration/profiles/patches.mdx
```diff
@@ -115,13 +118,19 @@ profiles:
   - op: replace
     path: images.backend.image
     value: john/stagingbackend
+  - op: replace
+    path: deployments.name=backend.helm.values.containers[0].image
```
LukasGentele, line 122: See comment below

docs/pages/configuration/profiles/patches.mdx
```diff
@@ -74,12 +74,15 @@ Using `op: add` only works as expected when `path` points to an array value. Usi
...
 
-1. Use the index of the array item you want to patch, e.g. `deployments[0]`
+1. Use the index of the array item you want to patch, e.g. `deployments[0]`, `deployments/0`
 2. Use a property selector matching the array item(s) you want to patch, e.g. `deployments.name=backend`
+3. Use a wildcard selector to match all array item(s), e.g. `deployments.*`,`deployments[*]` or `deployments/*`
+4. Use a comparison selector matching the array item(s) you want to patch, e.g. `deployments[?(@.helm.timeout > 1000)]`
+5. Use a regular expression selector matching the array item(s) you want to patch, e.g. `deployments[?(@.name=~/^production/)]`
 
-Using a property selector is often better because it is more resilient and will not cause any issues even if the order of an array's items is changed later on. A property selector is also able to select multiple array items if all of them have the same value for this property.
+Using a selector is often better because it is more resilient and will not cause any issues even if the order of an array's items is changed later on. A selector is also able to select multiple array items if all of them have the same value for this property.
```
LukasGentele, line 85: Using a selector *rather than the array index*...

LukasGentele (changes requested):

docs/pages/configuration/profiles/patches.mdx
lizardruss: I've update the examples after the selectors recommendation to use `deployments.name=backend.helm.values.containers.image=john/devbackend.image` or similar instead of indices.

Please have another look.

LukasGentele (approved):

PR #1666, opened 2021-09-20
feat: add option to execute hooks once

lizardruss:
### Changes
- Add new option: `hooks[*].container.once` that will configure remote container command hooks to run only once per deployment

### DIFF
--- a/docs/pages/configuration/hooks/basics.mdx
+++ b/docs/pages/configuration/hooks/basics.mdx
@@ -237,7 +237,7 @@ hooks:
 
 ## Execute hooks once
 
-Hooks can be executed once per deployment when executing a command in a container. This can be useful for running one-time development tasks without init containers. In the following example, the command would only run once in in all containers running with the image `nginx:1.21`.
+Hooks can be executed only once for each targeted container. This means as long as the container where the hook was executed stays running, the hook will not be executed for this container again until it restarts. This can be useful for running one-time development tasks without using init containers. In the following example, the command would only run once for the newest container running with the image `nginx:1.21`.
 
 ```yaml {8}
 ...
@@ -251,6 +251,8 @@ hooks:
   events: ["after:deploy"]
 ...
 ```
+
+If you run `devspace dev` or `devspace deploy` now multiple times and the container is not replaced or restarted, the hook is only executed once.
 
 ## Hook Context Information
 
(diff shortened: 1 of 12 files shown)


FabianKramm (approved):
@lizardruss thanks a lot for the PR! Looks good to me!

FabianKramm:
@LukasGentele please take a look at the docs

docs/pages/configuration/hooks/basics.mdx
```diff
@@ -235,6 +235,23 @@ hooks:
   events: ["after:build"]
 ```
 
+## Execute hooks once
+
+Hooks can be executed once per deployment when executing a command in a container. This can be useful for running one-time development tasks without init containers. In the following example, the command would only run once in containers matching the `nginx` image selector.
```
LukasGentele, line 240: "in containers matching the `nginx` image selector" => "in all containers running with the image `nginx`"

Ideally, we should add a tag or use a different image than `nginx` to make it easier to understand because this implies tag `latest` which is not a best practice

docs/pages/fragments/config-hooks.mdx
```diff
@@ -31,4 +31,5 @@ hooks:                              # struct[]  | Array of hooks to be executed.
     labelSelector: ...              # struct    | Key Value map of labels and values to select pods with
     containerName: ""               # string    | Container name to use after selecting a pod
     namespace: ""                   # string    | Kubernetes namespace to select pods in
+    once: ""                        # bool      | If true DevSpace will only execute the command once per deployment (default false)
```
LukasGentele, line 34: What does per deployment mean? Everytime the pipeline runs though? Maybe: `once - unless the targeted container is restarted which would allow re-executing this hook`? Not easy to express this accurately tbh

LukasGentele (changes requested):

docs/pages/configuration/hooks/basics.mdx
lizardruss: Changed to `nginx:1.21` to avoid implying `latest`

docs/pages/fragments/config-hooks.mdx
lizardruss: Re-worded to `If true the command will execute once - until the container is re-deployed using DevSpace (default false)`.

PR #1689, opened 2021-09-27
feat: add tolerations option to kaniko builds

### LINKED ISSUE #1399: Feature Request: Ability to define tolerations in kaniko builds
Currently, building images with kaniko allows to define a `nodeSelector` to build images on a specific Kubernetes cluster node. In some cases these nodes have more resources than others to perform that build tasks. To prevent any Pod to be scheduled on these nodes they usually have Taints and only Pods with the appropiate Tolerations are allowed to be scheduled.

A new `tolerations` option in `images.backend.build.kaniko` will be very helpful with `nodeSelector` to run builds on specific nodes with Taints

**Additional context**  
In our case, a node without Taints dedicated to builds is not an option.

/kind feature

asanCar:
/kind feature
/kind documentation

resolves #1399

* Added tolerations option to KanikoConfig struct
* Added new docs

No


FabianKramm:
@asanCar thanks a lot for the PR! LGTM!

PR #1735, opened 2021-10-20
Docker Compose in devspace init (conversion)

lizardruss:
### Changes
- detects `docker-compose.yaml` when running `devspace init` and provides an option to generate an initial devspace.yaml from docker compose
- Simulates docker compose volume behavior by uploading local files to an init container with the relevant volumes mounted. Once uploads are complete the init container exits and the application continues as expected.


FabianKramm (approved):

PR #1745, opened 2021-10-28
fix: local image lookup does not consider all tags

lizardruss:
### Changes
- Improve local image lookup to consider all tags


FabianKramm (approved):
@lizardruss thanks a lot for the PR! Looks good to me!

Issue #1743, opened 2021-10-28
File sync not working on macOS case insensitive file system


ragnarpa:
**What happened?**  

File sync not working on macOS case insensitive file system.

**What did you expect to happen instead?**  

File sync working on macOS regardless of case sensitivity.

**How can we reproduce the bug?** (as minimally and precisely as possible)  

https://github.com/ragnarpa/DevspacePathLetterCaseBug

**Local Environment:**  
- DevSpace Version: 5.16.2
- Operating System: mac
- Deployment method: helm

**Kubernetes Cluster:**  
- Cloud Provider: other
- Kubernetes Version: v1.21.5

**Anything else we need to know?**  

/kind bug

pratikjagrut:
fix is merged.

PR #1762, opened 2021-11-15
fix: adjust sync path for mac os

pratikjagrut:
/kind bugfix

resolves #
#1743

No

No

### DIFF
--- a/pkg/devspace/services/synccontroller/controller.go
+++ b/pkg/devspace/services/synccontroller/controller.go
@@ -218,6 +218,16 @@ func (c *controller) startWithWait(options *Options, log logpkg.Logger) error {
 
 	return nil
 }
+func realWorkDir() (string, error) {
+	if runtime.GOOS == "darwin" {
+		if pwd, present := os.LookupEnv("PWD"); present {
+			os.Unsetenv("PWD")
+			defer os.Setenv("PWD", pwd)
+		}
+		return os.Getwd()
+	}
+	return ".", nil
+}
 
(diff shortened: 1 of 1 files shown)


pkg/devspace/services/synccontroller/controller.go
```diff
@@ -226,10 +233,12 @@ func (c *controller) startSync(options *Options, onInitUploadDone chan struct{},
 	)
 
 	localPath := "."
+	if runtime.GOOS == "darwin" {
```
FabianKramm, line 236: Can you put the runtime.GOOS check inside the function?

pkg/devspace/services/synccontroller/controller.go
```diff
@@ -226,10 +233,12 @@ func (c *controller) startSync(options *Options, onInitUploadDone chan struct{},
...
+		localPath, _ = realWorkDir()
```
FabianKramm, line 237: Either return the error, or make sure the local path is "." again. Please also move this logic into the function

FabianKramm (changes requested):
@pratikjagrut thanks a lot for the PR! Please see the comments below

pkg/devspace/services/synccontroller/controller.go
pratikjagrut: done

pkg/devspace/services/synccontroller/controller.go
pratikjagrut: done

FabianKramm (approved):
@pratikjagrut thanks! LGTM!

PR #1765, opened 2021-11-18
WIP: feat: ask context and namespace if changed

pratikjagrut:
/kind feature

resolves #
If the Kube context or the namespace is changed it prompts a question asking which context/namespace to use.

Deprecates `-s` flag

No

### DIFF
--- a/cmd/analyze.go
+++ b/cmd/analyze.go
@@ -86,8 +86,9 @@ func (cmd *AnalyzeCmd) RunAnalyze(f factory.Factory, cobraCmd *cobra.Command, ar
 		return err
 	}
 
-	// Print warning
-	err = client.PrintWarning(generatedConfig, cmd.NoWarn, false, log)
+	// If the current kube context or namespace is different than old,
+	// show warnings and reset kube client if necessary
+	client, err = client.CheckKubeContext(generatedConfig, cmd.NoWarn, log)
 	if err != nil {
 		return err
 	}
--- a/pkg/devspace/kubectl/client.go
+++ b/pkg/devspace/kubectl/client.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"sort"
 	"time"
 
@@ -18,6 +19,7 @@ import (
 	"github.com/loft-sh/devspace/pkg/util/kubeconfig"
 	"github.com/loft-sh/devspace/pkg/util/log"
 	"github.com/loft-sh/devspace/pkg/util/survey"
+	"github.com/loft-sh/devspace/pkg/util/terminal"
 
 	"github.com/mgutz/ansi"
 	"github.com/pkg/errors"
@@ -51,7 +53,7 @@ type Client interface {
 
 	// PrintWarning this function will print a warning if the generated config contains a different last kube context / namespace
 	// than the one that is used currently
-	PrintWarning(generatedConfig *generated.Config, noWarning, shouldWait bool, log log.Logger) error
+	CheckKubeContext(generatedConfig *generated.Config, noWarning bool, log log.Logger) (Client, error)
 
 	// Copies and extracts files into the container from the reader interface
 	CopyFromReader(pod *k8sv1.Pod, container, containerPath string, reader io.Reader) error
@@ -97,6 +99,8 @@ type Client interface {
 	// Returns true if in cluster kubernetes configuration is detected
 	IsInCluster() bool
 }
+
+var _, tty = terminal.SetupTTY(os.Stdin, os.Stdout)
 
 type client struct {
 	Client       kubernetes.Interface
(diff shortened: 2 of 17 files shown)


pkg/devspace/kubectl/client.go
```diff
@@ -204,44 +208,75 @@ func (client *client) IsInCluster() bool {
...
-				log.Infof("Or use the '%s' flag to ignore this warning", ansi.Color("--no-warn", "white+b"))
-				wait = true
-			} else if generatedConfig.GetActive().LastContext.Namespace != "" && generatedConfig.GetActive().LastContext.Namespace != client.namespace {
+				// if terminal is not interactive then return the same client
+				if !tty.IsTerminalIn() {
+					return client, nil
+				}
+
+				kc, err := log.Question(&survey.QuestionOptions{
+					Question:     "Which context do you want use?",
```
FabianKramm, line 234: "Which context do you want to use?"

pkg/devspace/kubectl/client.go
```diff
@@ -204,44 +208,75 @@ func (client *client) IsInCluster() bool {
...
+					return client, nil
+				}
 
-			if wait && shouldWait {
-				log.StartWait("Will continue in 10 seconds...")
-				time.Sleep(10 * time.Second)
-				log.StopWait()
-				log.WriteString("\n")
+				ns, err := log.Question(&survey.QuestionOptions{
+					Question:     "Which namespace do you want use?",
```
FabianKramm, line 263: "Which namespace do you want to use?"

pkg/devspace/kubectl/client.go
```diff
@@ -250,10 +285,13 @@ func (client *client) PrintWarning(generatedConfig *generated.Config, noWarning,
...
 
 	// Info messages
-	log.Infof("Using namespace '%s'", ansi.Color(client.namespace, "white+b"))
-	log.Infof("Using kube context '%s'", ansi.Color(client.currentContext, "white+b"))
+	log.Infof("Using namespace '%s'", ansi.Color(namespace, "white+b"))
+	log.Infof("Using kube context '%s'", ansi.Color(kubeContext, "white+b"))
 
-	return nil
+	if resetClient {
+		return NewClientFromContext(kubeContext, namespace, false, client.kubeLoader)
```
FabianKramm, line 292: I think we should actually switch context here if the user decides to switch

pkg/devspace/kubectl/client.go
```diff
@@ -204,44 +208,75 @@ func (client *client) IsInCluster() bool {
...
+					Options: []string{
+						namespace,
+						generatedConfig.GetActive().LastContext.Namespace,
+					},
+				})
+				if err != nil {
+					return client, err
+				}
+				namespace = ns
+				resetClient = true
```
FabianKramm, line 274: Only reset if ns != namespace

pkg/devspace/kubectl/client.go
```diff
@@ -204,44 +208,75 @@ func (client *client) IsInCluster() bool {
...
+					Options: []string{
+						kubeContext,
+						generatedConfig.GetActive().LastContext.Context,
+					},
+				})
+				if err != nil {
+					return client, err
+				}
+				kubeContext = kc
+				resetClient = true
```
FabianKramm, line 245: Only reset if kc != kubeContext

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Please see comments below

LukasGentele:
Quick note on this one: I think we should not ask 2 questions. If the user gets asked to confirm the kube-context, we should not have to ask the namespace again.

pkg/devspace/kubectl/client.go
pratikjagrut: done.

pkg/devspace/kubectl/client.go
pratikjagrut: done.

pkg/devspace/kubectl/client.go
pratikjagrut: done.

pkg/devspace/kubectl/client.go
pratikjagrut: done.

pkg/devspace/kubectl/client.go
pratikjagrut: done.

pkg/devspace/kubectl/client.go
pratikjagrut: We're only reseting the client if user selects different context or namespace when asked.

pratikjagrut:
> Quick note on this one: I think we should not ask 2 questions. If the user gets asked to confirm the kube-context, we should not have to ask the namespace again.

Good point. Thanks.

pkg/devspace/kubectl/client.go
FabianKramm: @pratikjagrut yes, but the problem would be that the user would get asked this question everytime after the question, e.g.:
```
devspace dev -n test
devspace use namespace test2
devspace dev # <- question here, I select test
devspace dev # <- question here again etc.

```

FabianKramm (changes requested):

cmd/analyze.go
```diff
@@ -86,8 +89,11 @@ func (cmd *AnalyzeCmd) RunAnalyze(f factory.Factory, cobraCmd *cobra.Command, ar
 		return err
 	}
 
-	// Print warning
-	err = client.PrintWarning(generatedConfig, cmd.NoWarn, false, log)
+	// If the current kube context or namespace is different than old,
+	// show warnings and reset kube client if necessary
+	_, tty := terminal.SetupTTY(os.Stdin, os.Stdout)
```
FabianKramm, line 94: instead of initializing this everytime, please create a global variable in the kubectl package that you'll overwrite during tests

pkg/devspace/kubectl/client.go
```diff
@@ -250,10 +289,13 @@ func (client *client) PrintWarning(generatedConfig *generated.Config, noWarning,
...
 
 	// Info messages
-	log.Infof("Using namespace '%s'", ansi.Color(client.namespace, "white+b"))
-	log.Infof("Using kube context '%s'", ansi.Color(client.currentContext, "white+b"))
+	log.Infof("Using namespace '%s'", ansi.Color(currentConfigContext.Namespace, "white+b"))
+	log.Infof("Using kube context '%s'", ansi.Color(currentConfigContext.Context, "white+b"))
 
-	return nil
+	if resetClient {
+		return NewClientFromContext(currentConfigContext.Context, currentConfigContext.Namespace, false, client.kubeLoader)
```
FabianKramm, line 296: switch context here please

FabianKramm (changes requested):
@pratikjagrut please see the comments, besides that it looks good!

cmd/analyze.go
pratikjagrut: done

pkg/devspace/kubectl/client.go
pratikjagrut: done

FabianKramm (approved):
@pratikjagrut thanks! LGTM!

PR #1770, opened 2021-11-21
all: change map[string]bool to map[string]struct{} for performance gain

tranngoclam:
/kind enhancement

resolves #
Change map[string]bool to map[string]struct{} for performance gain

No

No

N/A

### DIFF
--- a/cmd/set/var.go
+++ b/cmd/set/var.go
@@ -80,7 +80,7 @@ func (cmd *varCmd) RunSetVar(f factory.Factory, cobraCmd *cobra.Command, args []
 			return errors.Errorf("Unexpected variable format. Expected key=value, got %s", v)
 		} else if variable.IsPredefinedVariable(splitted[0]) {
 			return errors.Errorf("cannot set predefined variable %s", splitted[0])
-		} else if !variableParser.Used[splitted[0]] {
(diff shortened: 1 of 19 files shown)


LukasGentele:
Thanks for contributing. I'd argue we should not merge this though because it is a micro optimization and performance is negligible in this case. We may do a boolean comparison 10-20 times at most for execution and how much time is it going to save to make this a tiny bit faster? 0.003 seconds? I doubt that it's of any significance.

However, making such a change really hurts code readability and maintainability, so definitely a clear downside to a tiny performance gain. Generally, making such micro optimizations should be the folks responsibility who work on the language and its compiler, not a high level project. If we were working on machine learning algorithms with millions of boolean comparisons in maps, I'd be agreeing with you that performance should be a priority over readability, but that's not the case here. 

I hope that helps understand my reasoning in opposing this change :) I still appreciate you looking into this. Happy to point you other topics that you may be able to work on if youd like to contribute to the project.

tranngoclam:
@LukasGentele
Thanks for sharing thoughts.

This may impact to the current code readability and not significantly an improvement for performance, I agree this point.
However, what do you think if we should bring this for future codebase, especially in some utility packages ?

Btw, I would like to know the roadmap of devspace which I can look into more deeply, appreciate if you can share this to me then I can have better knowledge about this project.

PR #1795, opened 2021-12-02
feat: add run commands to help usage

lizardruss:
### Changes
- Adds `devspace.yaml` configured commands to cobra command configuration.

### Details
`devspace --help` displays the custom commands under `Additional run commands:` using a custom template
```
DevSpace accelerates developing, deploying and debugging applications with Docker and Kubernetes. Get started by running the init command in one of your projects:

                devspace init

Usage:
  devspace [command]
  
Available Commands:
  add         Convenience command: adds something to devspace.yaml
  analyze     Analyzes a kubernetes namespace and checks for potential problems
  attach      Attaches to a container
  build       Builds all defined images and pushes them
  cleanup     Cleans up resources
  completion  Output shell completion for the given shell (bash or zsh)
  deploy      Deploy the project
  dev         Starts the development mode
  enter       Open a shell to a container
  help        Help about any command
  init        Initializes DevSpace in the current folder
  list        Lists configuration
  logs        Prints the logs of a pod and attaches to it
  open        Opens the space in the browser
  print       Print displays the configuration
  purge       Delete deployed resources
  remove      Changes devspace configuration
  render      Render builds all defined images and shows the yamls that would be deployed
  reset       Resets an cluster token
  restart     Restarts containers where the sync restart helper is injected
  restore     Restore configuration
  run         Run executes a predefined command
  save        Save configuration
  set         Make global configuration changes
  sync        Starts a bi-directional sync between the target container and the local path
  ui          Opens the localhost UI in the browser
  update      Updates the current config
  upgrade     Upgrade the DevSpace CLI to the newest version
  use         Use specific config
  
Flags:
      --config string                The devspace config file to use
      --debug                        Prints the stack trace if an error occurs
      --disable-profile-activation   If true will ignore all profile activations
  -h, --help                         help for devspace
      --inactivity-timeout int       Minutes the current user is inactive (no mouse or keyboard interaction) until DevSpace will exit automatically. 0 to disable. Only supported on windows and mac operating systems (default 180)
      --kube-context string          The kubernetes context to use
  -n, --namespace string             The kubernetes namespace to use
      --no-warn                      If true does not show any warning when deploying into a different namespace or kube-context than before
  -p, --profile strings              The DevSpace profiles to apply. Multiple profiles are applied in the order they are specified
      --profile-parent strings       One or more profiles that should be applied before the specified profile (e.g. devspace dev --profile-parent=base1 --profile-parent=base2 --profile=my-profile)
      --profile-refresh              If true will pull and re-download profile parent sources
      --restore-vars                 If true will restore the variables from kubernetes before loading the config
      --save-vars                    If true will save the variables to kubernetes after loading the config
      --silent                       Run in silent mode and prevents any devspace log output except panics & fatals
  -s, --switch-context               DEPRECATED: Switches and uses the last kube context and namespace that was used to deploy the DevSpace project
      --var strings                  Variables to override during execution (e.g. --var=MYVAR=MYVALUE)
      --vars-secret string           The secret to restore/save the variables from/to, if --restore-vars or --save-vars is enabled (default "devspace-vars")

Additional run commands:
  devspace run test1 Runs echo -n
  devspace run test2 Runs echo -n $@

Use "devspace [command] --help" for more information about a command.
```

`devspace run --help` displays the commands under `Available commands:` using using cobra's default behavior
```
#######################################################
##################### devspace run ####################
#######################################################
Run executes a predefined command from the devspace.yaml

Examples:
devspace run mycommand --myarg 123
devspace run mycommand2 1 2 3
devspace --dependency my-dependency run any-command --any-command-flag
#######################################################

Usage:
  devspace run [flags]
  devspace run [command]
  
Available Commands:
  test1       Runs echo -n
  test2       Runs echo -n $@
  
Flags:
      --dependency string   Run a command from a specific dependency
  -h, --help                help for run
  
Global Flags:
      --config string                The devspace config file to use
      --debug                        Prints the stack trace if an error occurs
      --disable-profile-activation   If true will ignore all profile activations
      --inactivity-timeout int       Minutes the current user is inactive (no mouse or keyboard interaction) until DevSpace will exit automatically. 0 to disable. Only supported on windows and mac operating systems (default 180)
      --kube-context string          The kubernetes context to use
  -n, --namespace string             The kubernetes namespace to use
      --no-warn                      If true does not show any warning when deploying into a different namespace or kube-context than before
  -p, --profile strings              The DevSpace profiles to apply. Multiple profiles are applied in the order they are specified
      --profile-parent strings       One or more profiles that should be applied before the specified profile (e.g. devspace dev --profile-parent=base1 --profile-parent=base2 --profile=my-profile)
      --profile-refresh              If true will pull and re-download profile parent sources
      --restore-vars                 If true will restore the variables from kubernetes before loading the config
      --save-vars                    If true will save the variables to kubernetes after loading the config
      --silent                       Run in silent mode and prevents any devspace log output except panics & fatals
  -s, --switch-context               DEPRECATED: Switches and uses the last kube context and namespace that was used to deploy the DevSpace project
      --var strings                  Variables to override during execution (e.g. --var=MYVAR=MYVALUE)
      --vars-secret string           The secret to restore/save the variables from/to, if --restore-vars or --save-vars is enabled (default "devspace-vars")

Use "devspace run [command] --help" for more information about a command.
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -164,6 +164,43 @@ func BuildRoot(f factory.Factory, excludePlugins bool) *cobra.Command {
 	persistentFlags := rootCmd.PersistentFlags()
 	globalFlags = flags.SetGlobalFlags(persistentFlags)
 
+	rootCmd.SetUsageTemplate(`Usage:{{if .Runnable}}
+  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}
+  
+Aliases:
+  {{.NameAndAliases}}{{end}}{{if .HasExample}}
+  
+Examples:
+{{.Example}}{{end}}{{if .HasAvailableSubCommands}}
+  
+Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
+  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}
+  
+Flags:
+{{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}
+  
+Global Flags:
+{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}
+
+Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
+  {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}
+
+{{- if (and .HasAvailableSubCommands) -}}
+{{- range .Commands -}}
+{{- if (and .HasSubCommands (eq .Name "run"))}}
+
+Additional run commands:
+{{- range .Commands}}
+  {{rpad .CommandPath .CommandPathPadding}} {{.Short}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end}}
+
+Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
+`)
+
 	// Add sub commands
 	rootCmd.AddCommand(add.NewAddCmd(f, globalFlags, plugins))
 	rootCmd.AddCommand(cleanup.NewCleanupCmd(f, globalFlags, plugins))
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -104,10 +104,11 @@ devspace --dependency my-dependency run any-command --any-command-flag
 			description = "Runs " + command.Command
 		}
 		runCmd.AddCommand(&cobra.Command{
-			Use:   command.Name,
-			Short: description,
-			Long:  description,
-			Args:  cobra.ArbitraryArgs,
+			Use:                command.Name,
+			Short:              description,
+			Long:               description,
+			Args:               cobra.ArbitraryArgs,
+			DisableFlagParsing: true,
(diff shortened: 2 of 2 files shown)


LukasGentele:
@lizardruss Thanks for implementing this so quickly! Ideally a `devspace run -h / --help` should show the command list as well. I understand why natively that is the same as `devspace run [command] -h / --help` but we need to differentiate here, unfortunately. I know that cobra as a framework may not help you with determining the difference but in that case, we need to (for this particular command) disable their args parsing and do it manually instead.

Biggest issue I see apart from the above: `devspace run test1 --foo now displays [fatal] unknown flag: --foo` => this breaking change is not acceptable, a lot of custom commands are defined that way, e.g. `command: devspace dev -p [some-profile]` should still allow `devspace run [command-name] -b` and lead to `devspace dev -p [profile-name] -b` execution. Anything else would be a huge issue for a lot of devspace users

lizardruss:
Pushed up a fix for the flag parsing.

FabianKramm (approved):

cmd/root.go
```diff
@@ -164,6 +164,43 @@ func BuildRoot(f factory.Factory, excludePlugins bool) *cobra.Command {
...
+
+Additional run commands:
+{{- range .Commands}}
+  {{rpad (printf "'%s'" .CommandPath) .CommandPathPadding}} {{.Short}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end}}
+
+Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
```
FabianKramm, line 201: I'm not sure if this might be confusing as people would expect that you can run `devspace run my-command --help`. Maybe we should remove this or move it above the `Additional run commands` section

PR #1803, opened 2021-12-03
fix: profiles should not activate on substring matches

lizardruss:
### Changes
- Prevents substring matches from activating profiles. For example: an expression `test` should not match `test123`

### DIFF
--- a/e2e/tests/config/testdata/profile-activation/regexp-exact.yaml
+++ b/e2e/tests/config/testdata/profile-activation/regexp-exact.yaml
@@ -1,9 +1,9 @@
-version: v1beta10
+version: v1beta11
 profiles:
   - name: one
     activation:
       - env:
-          FOO: "\\^t.*\\$"
+          FOO: "^t.*$"
     patches:
       - op: replace
         path: deployments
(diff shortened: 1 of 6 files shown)


e2e/tests/config/testdata/profile-activation/regexp-exact.yaml
```diff
@@ -0,0 +1,18 @@
+version: v1beta10
```
pratikjagrut, line 1: Shouldn't this be v1beta11?

pratikjagrut, line 1: Shouldn't this be v1beta11?

e2e/tests/config/testdata/profile-activation/regexp-substring.yaml
```diff
@@ -0,0 +1,18 @@
+version: v1beta10
```
pratikjagrut, line 1: Shouldn't this be v1beta11?

e2e/tests/config/testdata/profile-activation/string-exact.yaml
```diff
@@ -0,0 +1,18 @@
+version: v1beta10
```
pratikjagrut, line 1: Shouldn't this be v1beta11?

e2e/tests/config/testdata/profile-activation/regexp-exact.yaml
FabianKramm: Yeah would be good to use the newest config version here

FabianKramm (approved):

PR #1808, opened 2021-12-07
feat: allow variables and expression in profiles

lizardruss:
### Changes
- Allow profiles to be defined by expressions:
```yaml
profiles: ${cat profiles.yaml}
```

- Allow items of profiles array to be defined by expressions:
```yaml
profiles:
- ${cat profile1.yaml}
```

- Allow profile content to be defined by expressions:
```yaml
profiles:
- name: production
  merge:
    deployments: ${cat production.yaml}
```

- Validate that `name`, `parent`, `parents`, `activations`, `patches[*].path`, and `patches[*].op` are not variables or expressions
- Auto-activation of profiles using `vars` matching
- Documented where config expressions are not supported

### Optional TODOs ?
- [ ] move some tests from loader_test.go to version_test.go? profile activation logic is currently in versions.go, which is light on tests
- [ ] move some config e2e tests to loader_test.go. Many of these e2e tests could be implemented as unit tests.


FabianKramm (approved):

Issue #1813, opened 2021-12-11
Deployment does not work with kubectl 1.23.0


toddq:
**What happened?**  
Using devspace deployment with kubectl results in 

```
[fatal]  error deploying: error deploying server: error: --dry-run flag without a value was specified. A value must be set: "none", "server", or "client".
```

`kubectl --dry-run` has been deprecated awhile and was removed in kubectl 1.23.0 - 
https://github.com/kubernetes/kubernetes/pull/105327

**What did you expect to happen instead?**  

Deployment to at least be attempted.

**How can we reproduce the bug?** (as minimally and precisely as possible)  

- Install kubectl 1.23.0
- Configure devspace with a kubectl deployment manifest
- Run `devspace dev`

**Local Environment:**  
- DevSpace Version: 5.17.0
- Operating System:  mac
- Deployment method: kubectl apply

**Kubernetes Cluster:**  
- Cloud Provider: google | aws | azure | DevSpace Cloud | DevSpace Cloud Enterprise | other
- Kubernetes Version: [use `kubectl version`]

**Anything else we need to know?**  

Workaround - I was able to roll back my `kubernetes-cli` version to 1.22.4 (not easy with Homebrew) and things work as expected now.

/kind bug

pratikjagrut:
@toddq Thanks for creating this issue. Yes, kubectl has deprecated the --dry-run. We'll fix that with the next release

pratikjagrut:
Hi @toddq we've released the fix for this issue in v5.18.0. Can you please check and let us know if we can close this issue?

toddq:
Works great, thank you!  Close away.

PR #1814, opened 2021-12-13
fix: adjust --dry-run flag value depending on version of kubectl

pratikjagrut:
/kind bugfix
/kind enhancement

resolves #
#1813 

No

Yes


pkg/util/downloader/commands/kubectl.go
```diff
@@ -1,18 +1,33 @@
...
-	"strings"
 )
 
 var (
-	kubectlVersion  = "v1.21.2"
-	kubectlDownload = "https://storage.googleapis.com/kubernetes-release/release/" + kubectlVersion + "/bin/" + runtime.GOOS + "/" + runtime.GOARCH + "/kubectl"
+	kubectlVersion = func() string {
+		res, err := http.Get("https://storage.googleapis.com/kubernetes-release/release/stable.txt")
+		if err != nil {
+			log.Fatal(err)
```
FabianKramm, line 21: we shouldn't fatal here as this will break devspace completely if this is not reachable anymore

pkg/util/downloader/commands/kubectl.go
```diff
@@ -1,18 +1,33 @@
...
+		}
+		content, err := ioutil.ReadAll(res.Body)
+		res.Body.Close()
+		if err != nil {
+			log.Fatal(err)
```
FabianKramm, line 26: same here

pkg/util/downloader/commands/kubectl.go
```diff
@@ -1,18 +1,33 @@
...
 	"github.com/otiai10/copy"
-	"path/filepath"
-	"runtime"
-	"strings"
 )
 
 var (
-	kubectlVersion  = "v1.21.2"
-	kubectlDownload = "https://storage.googleapis.com/kubernetes-release/release/" + kubectlVersion + "/bin/" + runtime.GOOS + "/" + runtime.GOARCH + "/kubectl"
+	kubectlVersion = func() string {
```
FabianKramm, line 18: move this functionality into a code block that will be executed only if this version is needed

pkg/devspace/deploy/deployer/kubectl/builder.go
```diff
@@ -81,7 +83,30 @@ func (k *kubectlBuilder) Build(manifest string, cmd RunCommand) ([]*unstructured
...
 
-	args = append(args, "--dry-run", "--output", "yaml", "--validate=false")
+	// compare kubectl version for --dry-run flag value
+	out, err := command.NewStreamCommand(k.path, []string{"version", "--client", "--short"}).Output()
+	if err != nil {
+		return nil, err
+	}
+
+	versionSlice := strings.Split(string(out), "v")
+	v1, err := version.NewVersion(strings.TrimSpace(versionSlice[1]))
```
FabianKramm, line 93: use `strings.TrimPrefix(string(out), "v")` here instead of split & trimSpace

FabianKramm (changes requested):
@pratikjagrut thanks for this PR! Please take a look at my comments

pkg/devspace/deploy/deployer/kubectl/builder.go
pratikjagrut: This way is not working. Not sure why but it is not returning a trimmed value.

pratikjagrut: And there're trailing space in out so anyhow we'll need to use trimSpaces.

FabianKramm: Then use `version.NewVersion(strings.TrimPrefix(strings.TrimSpace(string(out)), "Client Version: v"))`, the problem with the split approach is that you expect that there will be a v..., but there might be situations where this is not the case, for example if you self-compile kubectl or if they change the way the version is printed, we shouldn't panic here

FabianKramm (changes requested):

FabianKramm (approved):
LGTM! Thanks @pratikjagrut

PR #1815, opened 2021-12-14
fix: use the given binary path when finishing the install with npm/yarn

lizardruss:
### Changes
- add support for arm64 architecture
- use the given binary directory when finalizing install


FabianKramm (approved):

PR #1819, opened 2021-12-15
Fix Alpine / Busybox compatibility

### LINKED ISSUE #1185: Installation of devspace inside node:alpine container fails
**What happened?**  
I am trying to run devspace commands inside a gitlab pipeline. For that i need to install devspace inside a docker container with image `node:alpine`. After installation when i run `devspace -h` it should usually pull some binaries and then present the help. Instead console is flooded.

I am able to install devspace successful inside `node:alpine`

```bash
docker run -ti --entrypoint sh node:alpine
```

Then inside the container:

```bash
npm -g install devspace
# installation OK
devspace -h
```

Console is flooded:

```bash
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
sh: out of range
```

**Local Environment:**  
- docker v19
- node:alpine

/kind bug

stickeegreg:
/kind bugfix

resolves #1185 

Fixed an issue where DevSpace does not run in Alpine when installed via npm

### DIFF
--- a/dist/npm/bin/devspace
+++ b/dist/npm/bin/devspace
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 
 BINARY=$(command -v devspace)
-STAT=$(stat -L -f%z "$BINARY" 2>/dev/null || stat --printf="%s" "$BINARY" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


FabianKramm (approved):
@stickeegreg thanks for creating this PR! LGTM!

Issue #1818, opened 2021-12-15
Invalid memory address or nil pointer dereference


bforchhammer:
**What happened?**  

After upgrading devspace to version 5.18.0 many commands, e.g. `devspace dev` or `devspace deploy`, error with the following message:

```
❯ devspace dev
[info]   Using namespace 'test'
[info]   Using kube context 'docker-desktop'
[warn]   Cyclic dependency found:
...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x5402860]

goroutine 1 [running]:
github.com/loft-sh/devspace/pkg/devspace/dependency.(*Dependency).Name(0x56bf680)
        /Users/runner/work/devspace/devspace/pkg/devspace/dependency/dependency.go:425
github.com/loft-sh/devspace/pkg/devspace/dependency.(*manager).executeDependenciesRecursive(0xc000124440, {0xc000dbb110, 0xc000677400}, {0xc000677400, 0x5, 0x4400ce5}, {0x0, 0x0, 0x0}, {0x0, ...}, ...)
        /Users/runner/work/devspace/devspace/pkg/devspace/dependency/dependency.go:271 +0x1b0
github.com/loft-sh/devspace/pkg/devspace/dependency.(*manager).executeDependenciesRecursive(0xc000124440, {0xc0003a5b58, 0xc000677480}, {0xc000677480, 0x6, 0x0}, {0x0, 0x0, 0x0}, {0x0, ...}, ...)
        /Users/runner/work/devspace/devspace/pkg/devspace/dependency/dependency.go:283 +0x451
github.com/loft-sh/devspace/pkg/devspace/dependency.(*manager).executeDependenciesRecursive(0xc000124440, {0x0, 0x0}, {0xc0011a2600, 0xb, 0x52}, {0x0, 0x0, 0x0}, {0x0, ...}, ...)
        /Users/runner/work/devspace/devspace/pkg/devspace/dependency/dependency.go:283 +0x451
github.com/loft-sh/devspace/pkg/devspace/dependency.(*manager).handleDependencies(0xc000124440, {0x0, 0x0, 0x0}, {0x0, 0x0, 0x0}, 0x30, 0x1a, 0x1, ...)
        /Users/runner/work/devspace/devspace/pkg/devspace/dependency/dependency.go:235 +0x471
[...]

PR #1822, opened 2021-12-15
fix: nil cyclic dependency error

pratikjagrut:
/kind bugfix

resolves #
#1818 

Fixed an issue where DevSpace was getting `Invalid memory address or nil pointer dereference` while resolving cyclic dependency. Also adding an e2e test for the same bug.


FabianKramm (approved):
@pratikjagrut thanks for the PR! LGTM!

PR #1824, opened 2021-12-16
refactor: make workflows independent and minimize the unnecessary runs

pratikjagrut:
/kind enhancement

resolves #
This PR minimizes the unnecessary runs of ci.


.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,55 @@
+name: Unit test
+
+on:
+  release:
+    types: [created]
+  pull_request:
+    branches:
+      - main
```
FabianKramm, line 8: we still use master branch, so I guess we need to change that here

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,27 @@
+name: Lint
+
+on:
+  release:
+    types: [created]
+  pull_request:
+    branches:
+      - main
```
FabianKramm, line 8: we still use master branch, so I guess we need to change that here

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,49 @@
+name: E2E tests
+
+on:
+  release:
+    types: [created]
+  pull_request:
+    branches:
+      - main
```
FabianKramm, line 8: we still use master branch, so I guess we need to change that here

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,55 @@
...
+          go-version: 1.17
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v1
+
+      - name: Test
+        run: ./hack/coverage.bash
+
+  test-unit-windows:
+    if: github.ref != 'refs/heads/master'
```
FabianKramm, line 40: why does we have this for windows but not for the others?

.github/workflows/pr-checks.yml
```diff
@@ -1,92 +0,0 @@
-name: PR checks
-
-on: 
-  pull_request:
-    branches: [ master ]
-
-env:
```
FabianKramm, line 7: are these not needed anymore if we split up those checks?

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Please see my comments

.github/workflows/unit-tests.yaml
pratikjagrut: I guess this is a mistake. We don't need any condition here.

.github/workflows/unit-tests.yaml
pratikjagrut: ah yes.

.github/workflows/pr-checks.yml
pratikjagrut: Yes, we don't need this file.

FabianKramm (approved):
thanks @pratikjagrut ! LGTM!

PR #1846, opened 2022-01-06
fix: allow matching of integers using older patch path syntax

lizardruss:
### Changes

- modify patch path conversion to account for possible integer types.

### Notes
- This does not account for floating point numbers, but that is unlikely to be needed


FabianKramm (approved):

Issue #1880, opened 2022-01-28
evicted pods are still showing up in the devspace commands


ansonlc:
**Is your feature request related to a problem?**
when a pod is evicted (not sure why the evicted pod didn’t get deleted after 24 h) we can still see it in the devspace enter or other commands (devspace logs etc.)

**Which solution do you suggest?**
can we add a filter to only run those commands against healthy pods? Or any way to delete evicted pods automatically?

**Which alternative solutions exist?**

**Additional context**

/kind feature

FabianKramm:
@ansonlc thanks a lot for creating this issue! Yes that makes a lot of sense, we'll exclude evicted pods from showing up in `devspace` commands

pratikjagrut:
@ansonlc this feature was released in [v5.18.3](https://github.com/loft-sh/devspace/releases/tag/v5.18.3) could you please try it out and close this issue or update this issue accordingly? Thank you.

PR #1882, opened 2022-01-31
feat: check if pod is evicated before listing it

pratikjagrut:
/kind feature

resolves #
#1880


FabianKramm (changes requested):
@pratikjagrut thanks for creating this PR! Currently this PR has the following problems:
1. It selects an evicted pod if there is only 1 pod in total
2. It does not work if a container should get selected as in `devspace enter` for example

pkg/devspace/kubectl/selector/selector.go
```diff
@@ -410,3 +417,12 @@ func byImageName(ctx context.Context, client kubectl.Client, namespace string, i
 	}
 	return retPods, nil
 }
+
+func isPodEvicted(pod *corev1.Pod) bool {
```
FabianKramm, line 421: simplify this function to:
```
func isPodEvicted(pod *corev1.Pod) bool {
  return strings.Contains(pod.Status.Reason, "Evicted")
}
```

FabianKramm (changes requested):
@pratikjagrut thanks for the update! Looks good to me, but please change the function like mentioned in the comment above

pkg/devspace/kubectl/selector/selector.go
pratikjagrut: done

FabianKramm (approved):

PR #1884, opened 2022-01-31
Feature/deployment concurrency

### LINKED ISSUE #1877: Concurrent deployments to reduce deploy times.
Many of our projects using DevSpace spin up deployments that are not dependent on each other that we would like to deploy concurrently, like how images are built, to reduce deploy times.

For example, we may have an project with the following deployments:
- postgres (using helm)
- redis (using helm)
- server (using helm)

In such a case, the postgres and redis deployments do not have any dependency on another deployment, while the server relies on both the cache and database. We have projects with 3x more independent deployments than in this example and would like to have the option to concurrently apply ***select deployments***.

A colleague (@kuuji ) and I were hoping to make a contribution to the project to allow deployments to be applied concurrently following the example set by concurrent image builds with some differences.

The main difference is that we think there is value in specifying which deployments can be applied concurrently and another is that concurrency should be disabled by default to in order to maintain backwards compatibility.

For enabling concurrency on a deployment, we are suggesting an additional boolean on deployments like this:

```yaml
deployments:
- name: postgres
  helm: { ... }
  concurrency: true
- name: redis
  helm: { ... }
  concurrency: true
- name: server
  helm: { ... }
  concurrency: false # could be omitted as we suggest this value be false by default.
```

Since we are using helm for everything we could create a custom umbrella chart that wraps all our deployments that can be deployed concurrently and have that as the first DevSpace deployment in the list so that it is applied first. But this approach shifts more complexity to the user and supporting it feels in line with DevSpace's goal of "_develop[ing] and deploy[ing] cloud-native software faster_".

**Additional context**  
We are excited to work on this and would really appreciate feedback. Can't promise we won't get started testing something out before that, tho :P

/kind feature

erikkrieg:
/kind enhancement
/kind feature

Enables deployments to opt into deploying concurrently in order to improve performance for configs with many deployments.

resolves #1877

We did some tests with a modified version of `examples/quickstart` and found a roughly 35% - 50% reduction in deployment time. The more deployments, the greater the improvement.

- 3 concurrent vs 3 sequential deployments saw ~35% reduction in time.
- 6 concurrent vs 6 sequential deployments saw ~50% reduction in time.

Allow select deployments to deploy concurrently by setting `concurrent: true`.

The logging was changed to resemble how concurrent image build logs are handled. We attempted to capture the effect the sequential deploys get from using `StartWait` but this is something we were less certain about.

Example snippet of logging output:
```
[redis] Done: Successfully deployed redis with helm
[pgsql] Done: Deployed helm chart (Release revision: 1)
[pgsql] Done: Successfully deployed pgsql with helm
[wait] ⠋ Deploying 2 deployments concurrently (13s)
```
Something we didn't love but are not sure how to address is that as the number of remaining deployments decrements a new `StartWait` call is made and therefore the count of seconds starts back at 0.

We also wanted to wait for feedback before updating any documentation and will certainly do this if/when things are looking good :D


pratikjagrut:
@erikkrieg Thanks a lot for the PR! :)

pratikjagrut:
It'll also help if you could add an e2e test for this feature. You can refer to some existing e2e tests under `./e2e/tests/deploy` and add a new one to the same dir. @FabianKramm WDYT?

examples/quickstart-concurrent/.gitignore
```diff
@@ -0,0 +1,6 @@
+Dockerfile
+.devspace/
+chart/
+node_modules/
+test/
+devspace.yaml
```
pratikjagrut, line 6: Please remove devspace.yaml from .gitignore. It should be part of the example.

pratikjagrut (changes requested):

examples/quickstart-concurrent/.gitignore
kuuji: My bad, I'll clean that up. I just duplicated it from quickstart and didn't notice the gitignore included devspace.yaml.

kuuji:
Thanks for the suggestion @pratikjagrut 
I just added some e2e tests for both all concurrent and some concurrent some sequential. 

I also included some vscode config to run e2e, similar to what vcluster has. Let me know if you prefer that specific part to be in a different PR since it's not really related to this PR.

e2e/tests/deploy/deploy.go
```diff
@@ -194,4 +194,77 @@ var _ = DevSpaceDescribe("deploy", func() {
 		framework.ExpectNoError(err)
 		framework.ExpectEqual(out, "test")
 	})
+
+	ginkgo.It("should deploy helm applications concurrently", func() {
```
pratikjagrut, line 198: IMHO test case name should be renamed to "should deploy applications concurrently". Because this feature supposes to work with both kubectl and helm deployment configs.

e2e/tests/deploy/deploy.go
```diff
@@ -194,4 +194,77 @@ var _ = DevSpaceDescribe("deploy", func() {
...
+		out2, err := kubeClient.ExecByContainer("app.kubernetes.io/component=test-2", "container-0", ns, []string{"echo", "-n", "test"})
+		framework.ExpectNoError(err)
+		out3, err := kubeClient.ExecByContainer("app.kubernetes.io/component=test-3", "container-0", ns, []string{"echo", "-n", "test"})
+		framework.ExpectNoError(err)
+
+		framework.ExpectEqual(out, "test")
+		framework.ExpectEqual(out2, "test")
+		framework.ExpectEqual(out3, "test")
+	})
+	ginkgo.It("should deploy helm applications mixed concurrently and sequentially", func() {
```
pratikjagrut, line 234: Please add a new line here to separate both the test cases.

pratikjagrut, line 234: Also, the helm should be removed from the test case name for the same reason as above.

pratikjagrut (changes requested):
E2E tests LGTM just some minor changes.

examples/quickstart-concurrent/.gitignore
pratikjagrut: Thank you for resolving it.

pratikjagrut:
> I also included some vscode config to run e2e, similar to what vcluster has. Let me know if you prefer that specific part to be in a different PR since it's not really related to this PR.

Personally, I would have prefered separate PR for this change as it is not an intrinsic part of this feature.

examples/quickstart-concurrent/devspace_start.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
FabianKramm, line 1: please remove this file as its not really needed by the e2e test

examples/quickstart-concurrent/devspace.yaml
```diff
@@ -0,0 +1,69 @@
...
+        command: ["sleep", "infinity"]
+- name: quickstart-seq-1
+  helm:                              # This deployment uses `helm` but you can also define `kubectl` deployments or kustomizations
+    componentChart: true
+    displayOutput: true             # We are deploying the so-called Component Chart: https://devspace.sh/component-chart/docs
+    values:                          # Under `values` we can define the values for this Helm chart used during `helm install/upgrade`
+      containers:
+      - image: ${IMAGE}              # Use the value of our `${IMAGE}` variable here (see vars above)
+        command: ["sleep", "infinity"]
+dev:                                 # `dev` only applies when you run `devspace dev`
```
FabianKramm, line 59: remove this section please as its not used by the e2e test

examples/quickstart-concurrent/index.js
```diff
@@ -0,0 +1,27 @@
+const express = require('express')
```
FabianKramm, line 1: file also not used

examples/quickstart-concurrent/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
FabianKramm, line 1: package.json & package-lock.json are not used

FabianKramm (changes requested):
@erikkrieg @kuuji thanks so much for this PR! Thats really an improvement for DevSpace! Besides some small things in the e2e test the PR looks great and is ready to get merged!

erikkrieg:
Included documentation. Feeling pretty good about most of it but not sure whether the warning about an edge case with Helm hooks and concurrency is too specific or not.

FabianKramm (approved):
@erikkrieg thanks a lot, LGTM now!

erikkrieg:
Thanks, @FabianKramm and @pratikjagrut!

PR #1890, opened 2022-02-04
refactor: standardize image question across helm, kubectl, and kustomize

lizardruss:
### Changes
- Standardize image question for helm, kubectl and kustomizations deployments


FabianKramm (approved):

PR #1894, opened 2022-02-09
an attempt to revamp devspace-restart-helper

leventogut:
/kind enhancement

resolves #

Improve devspace-restart-helper:
- Full support for graceful exit (some fixes on responding signals.)
- Configurable Variables:
  --grace-period 7: Grace period -in seconds- to wait for a process to exit after sending its STOPSIGNAL, here we have several processes command, screen (if enabled), and some children of ours. The gracePeriod will be applied as one for each process. It is recommended to use 1/4, 1/5 of the terminationGracePeriodSeconds (default 30 seconds) value.
  --stop-signal-for-process (default 15/SIGTERM)
- better portability
- 

Fixed an issue where DevSpace restart helper was not responding SIGTERM under certain conditions.
Improved portability of DevSpace restart helper.
Improved flexibility of restart helper with configurable options.


FabianKramm:
@leventogut thanks a lot for the PR! The new helper looks great, but would it be possible to add an e2e test that tests this explicitly so that we can ensure in the pipeline that this works? If you don't know how to do that I'm happy to help anytime

PR #1942, opened 2022-03-15
feat: add get_config_value command

pratikjagrut:
/kind feature

### DIFF
--- a/pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
+++ b/pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
@@ -1,0 +1,54 @@
+package commands
+
+import (
+	"fmt"
+	"strings"
+
+	devspacecontext "github.com/loft-sh/devspace/pkg/devspace/context"
+	"github.com/vmware-labs/yaml-jsonpath/pkg/yamlpath"
+	"gopkg.in/yaml.v3"
+	"mvdan.cc/sh/v3/interp"
+)
+
+func GetConfigValue(ctx *devspacecontext.Context, args []string) error {
+	ctx.Log.Debugf("get_config_value %s", strings.Join(args, " "))
+	if len(args) != 1 {
+		return fmt.Errorf("usage: get_config_value deployments.my-deployment.helm.chart.name")
+	}
+
+	hc := interp.HandlerCtx(ctx.Context)
+	rawConfig := ctx.Config.Raw()
+
+	nodePath, err := yamlpath.NewPath(args[0])
+	if err != nil {
+		_, _ = hc.Stderr.Write([]byte(err.Error()))
+		return nil
+	}
+
+	out, err := yaml.Marshal(rawConfig)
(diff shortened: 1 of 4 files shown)


pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+
+	nodes, err := nodePath.Find(&doc)
+	if err != nil {
+		return err
+	}
+
+	hc := interp.HandlerCtx(ctx.Context)
+	_, _ = hc.Stdout.Write([]byte(nodes[0].Value))
```
FabianKramm, line 43: please check if there is len(nodes) > 0

pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
```diff
@@ -0,0 +1,56 @@
...
+	return nil
+}
+
+var endsWithBracketExp = regexp.MustCompile(`\[([^\]]+)\]$`)
+
+func getNodePath(path string) string {
+	if endsWithBracketExp.MatchString(path) {
+		return endsWithBracketExp.ReplaceAllString(path, "")
```
FabianKramm, line 51: this replaces `containers[0]` with `containers`, is that desired? What if I want to receive a value of `containers[1]` and not `containers[0]`?

pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	tokens := strings.Split(path, ".")
+	return strings.Join(tokens[0:], ".")
```
FabianKramm, line 55: why use `tokens[0:]` instead of `tokens`?

FabianKramm (changes requested):
@pratikjagrut thanks for the PR! Looks good, please see my comments below

pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
pratikjagrut: okay

pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
pratikjagrut: ah used it as it is from patches but I guess we can use tokens instead of tokens[0:]

pkg/devspace/pipeline/engine/pipelinehandler/commands/get_config_value.go
pratikjagrut: It will also work for containers[1]. I'll add one more testcase for this.

FabianKramm (approved):
@pratikjagrut thanks! LGTM!

PR #1958, opened 2022-03-23
fix: use notify library for symlinks

lizardruss:
### Changes
- Add e2e test for syncing symlinked directories
- Fix missing ssh argument
- Refactor symlink syncing to use notify library
- Fix warning about a "lost cancel" when returning errors (sync.go)


FabianKramm (approved):

PR #1963, opened 2022-03-23
feat: enable multiple pull secrets to be used by a service account

lizardruss:
### Changes
- Enable multiple pull secrets to used by a single service account


FabianKramm (approved):

PR #2002, opened 2022-03-28
fix: remove merged logs functionality to avoid UI errors

lizardruss:
### Changes
- fix a JavaScript error when loading a configuration with no profiles


FabianKramm (approved):

PR #2051, opened 2022-04-27
Added --set and --set-string flags to run_pipelines and run_default_pipeline commands along with e2e tests

tukobadnyanoba:
/kind enhancement
Added --set and --set-string flags to run_pipelines and run_default_pipeline commands
/kind test
Added e2e tests for both run_pipelines and run_default_pipeline new flags
resolves #
Two flags are now available for run_pipelines and run_default_pipeline.

NA

NA

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -1,12 +1,14 @@
 package framework
 
 import (
+	"context"
 	"fmt"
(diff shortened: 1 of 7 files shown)


FabianKramm (approved):
@tukobadnyanoba thanks a lot for the PR! LGTM!

PR #2061, opened 2022-05-02
Add unset nature to set and set string flags

tukobadnyanoba:
/kind enhancement
Sets the keys to null when they are added to flags.
resolves #


FabianKramm (approved):
@tukobadnyanoba thanks for the PR! LGTM!

FabianKramm (approved):
@tukobadnyanoba thanks for the PR! LGTM!

PR #2067, opened 2022-05-06
removed fake k8s client

tukobadnyanoba:
/kind bugfix
Earlier for non-existing kubeconfigs, we used to create fakeclientset, I have removed that logic now. If the kubeconfig is needed  for build operation using kaniko and buildkit, there are already checks present to validate. If the docker build is fired, we don't need k8s cluster, so its now executed without any error.

resolves #


FabianKramm (approved):
@tukobadnyanoba thanks a lot for the PR! LGTM!

PR #2086, opened 2022-05-26
Resolve all config paths at the same time for devspace print

mcastellini:
/kind enhancement

Relates to a possible improvement mentioned in https://github.com/loft-sh/devspace/issues/1829#issuecomment-995790591

Fixed an issue where DevSpace print would not resolve variables in under deployments.*.helm.values.


FabianKramm:
@mcastellini thanks a lot for the PR! LGTM!

PR #2093, opened 2022-06-06
added kind support

tukobadnyanoba:
/kind enhancement

resolves #

Fixed an issue where DevSpace ...now loads local images to kind cluster in case it is the kind context.

### DIFF
--- a/pkg/devspace/build/builder/buildkit/buildkit.go
+++ b/pkg/devspace/build/builder/buildkit/buildkit.go
@@ -204,8 +204,27 @@ func buildWithCLI(ctx context.Context, dir string, context io.Reader, writer io.
 			return fmt.Errorf("error retrieving minikube environment with 'minikube docker-env --shell none'. Try setting the option preferMinikube to false: %v", err)
 		}
 	}
+	err = command2.CommandWithEnv(ctx, dir, writer, writer, context, minikubeEnv, command[0], completeArgs...)
+	if err != nil {
+		return err
+	}
+	//load image if it is a kind-context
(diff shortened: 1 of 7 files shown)


pkg/devspace/build/builder/buildkit/buildkit.go
```diff
@@ -204,8 +205,28 @@ func buildWithCLI(ctx context.Context, dir string, context io.Reader, writer io.
...
+	if err != nil {
+		return err
+	}
+	//load image if it is a kind-context
+	if skipPush && kubeClient != nil && kubectl.IsKindContext(kubeClient.CurrentContext()) {
+		if len(options.Tags) > 0 {
+			for _, tag := range options.Tags {
+				var stdout, stderr bytes.Buffer
+				cmd := exec.Command("kind", "load", "docker-image", tag)
+				cmd.Stdout = &stdout
```
FabianKramm, line 218: I think we should use writer here as inside the other command. Besides it would be good to use the `command2` package here.

pkg/devspace/build/builder/buildkit/buildkit.go
```diff
@@ -204,8 +205,28 @@ func buildWithCLI(ctx context.Context, dir string, context io.Reader, writer io.
...
+				cmd.Stderr = &stderr
+				err := cmd.Run()
+				if err != nil {
+					return errors.Errorf("error during image load to kind cluster: %v", err)
```
FabianKramm, line 222: Emitting a warning here should be enough

FabianKramm (changes requested):
@tukobadnyanoba thanks for the PR! Please see my comments below

FabianKramm (approved):

PR #2180, opened 2022-07-14
made alwaysResolve default to true

### LINKED ISSUE #2164: vars not always populating as expected
**What happened?**  

I'm having some issues with variables not populating or even registering as expected.

For example, I created this extremely simple `devspace.yaml` file and ran `devspace list vars` and the variable does not show up. Note that `devspace print` does not show the vars either. 

```yaml
version: v2beta1
name: test
vars:
  VAR1: one
```

`devspace list vars` shows the below. 

```
    
         VARIABLE      |                       VALUE                        
  ---------------------+----------------------------------------------------
    DEVSPACE_CONTEXT   |                                                    
    DEVSPACE_NAME      | test                                               
    DEVSPACE_NAMESPACE |                                                    
    DEVSPACE_PROFILE   |                                                    
    DEVSPACE_RANDOM    | DLaoCo                                             
    DEVSPACE_TIMESTAMP | 1657206452                                         
    DEVSPACE_TMPDIR    | /var/folders/9z/25lc8jjn6mj7c4_yj7nyj1fw0000gp/T/  
    DEVSPACE_USER_HOME | /Users/tc                                          
    DEVSPACE_VERSION   | 6.0.0-beta.6                                       
    devspace.context   |                                                    
    devspace.namespace |               
```
In other cases, I have variables defined that ARE getting populated correctly, and they're defined right next to variables that are NOT getting populated correctly, which is super strange. 

Another thing I've noticed is that variables do not seem to be getting populated from my local environment when I don't provide a value for them. I can work around this by doing `AWS_USER: $(echo $AWS_USER)` or something similar, but I think the behavior is supposed to be that the variable will come from the local environment if it is not defined. Note that I wouldn't expect the variable to show up with `devspace list vars` if I don't list it as a VAR, but I would expect to be able to use it as `${AWS_USER}` in other places in my `devspace.yaml` file. 

This whole thing smells of a weird caching issue, but I haven't been able to track down any variables listed in any of the relevant `cache.yaml` files yet. 

I would expect all appropriate variables to show up when I run `devspace list vars`, or `devspace print`. 

**Local Environment:**  
- DevSpace Version: `devspace version 6.0.0-beta.6`
- Operating System: mac
**Kubernetes Cluster:**  
- Cloud Provider: aws 

/kind bug

tukobadnyanoba:
/kind enhancement

resolves #2164

Fixed an issue where DevSpace ...alwaysResolve set to true

### DIFF
--- a/pkg/devspace/config/versions/latest/schema.go
+++ b/pkg/devspace/config/versions/latest/schema.go
@@ -329,7 +329,7 @@ type DockerConfig struct {
 	// context is used.
 	PreferMinikube *bool `yaml:"preferMinikube,omitempty" json:"preferMinikube,omitempty" jsonschema_description:"PreferMinikube allows you to turn off using the minikube docker daemon if the minikube context is used."`
 	// UseCLI specifies if DevSpace should use the docker cli for building
-	UseCLI bool `yaml:"useCli,omitempty" json:"useCli,omitempty" jsonschema_description:"// UseCLI specifies if DevSpace should use the docker cli for building."`
+	UseCLI bool `yaml:"useCli,omitempty" json:"useCli,omitempty" jsonschema_description:"UseCLI specifies if DevSpace should use the docker cli for building."`
 	// Args are additional arguments to pass to the docker cli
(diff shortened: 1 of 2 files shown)


FabianKramm:
Closed in favor of #2183

PR #2182, opened 2022-07-15
fix: dev containers not being processed in nested format

lizardruss:
/kind bugfix

The following replaces the devImage correctly:
```yaml
dev:
  loft:
    labelSelector:
      app: loft
    devImage: loftsh/go
```

However the following format fails to replace the image with the devImage:
```yaml
dev:
  loft:
    labelSelector:
      app: loft
    containers:
      manager:
        devImage: loftsh/go
```

Fixed an issue where DevSpace would not process multiple named dev containers


FabianKramm (approved):

PR #2206, opened 2022-08-05
fix: op patching to not add elements if key already exists

carlmontanari:
/kind bugfix
/kind test

N/A

Fixed an issue where DevSpace add patch operations on existing keys breaks patching.

This fixes an issue where adding an already existing key to a map would basically append the value in your patch to the key. Unfortunately this would basically shift the yaml around so now the previous value would be the key on the next line. Example: 

```
   importme:
        build:
            buildKit: {}
        context: ../..
        veryimportantpath: dockerfile
        ./Dockerfile: image
        golang:1.17: rebuildStrategy
```

The above output is an example where there was a context set to "veryimportantpath", and an add op setting the context to "../..". This would obviously cause some errors. Moreover subsequent operations may try to reference the key "dockerfile" which is actually no longer a key in the document, causing an exception like `remove operation does not apply: doc is missing path: images.agent.dockerfile`. 

The `Perform` function got a little more complicated than I'd like to account for not breaking the ability to add elements into existing arrays.

### DIFF
--- a/pkg/devspace/config/loader/patch/operation.go
+++ b/pkg/devspace/config/loader/patch/operation.go
@@ -38,17 +38,14 @@ func (op *Operation) Perform(doc *yaml.Node) error {
 		return fmt.Errorf("%s operation does not apply: doc is missing path: %s", op.Op, op.Path)
 	}
 
-	var f func(parent *yaml.Node, match *yaml.Node)
+	// function that will actually perform the patch operation
+	var opFunc func(parent *yaml.Node, match *yaml.Node)
 
 	switch op.Op {
 	case opAdd:
-		f = op.add
-
-		var pathMatches bool
+		opFunc = op.add
 
 		if len(matches) > 0 {
-			pathMatches = true
-
 			if matches[0].Kind == yaml.MappingNode || matches[0].Kind == yaml.SequenceNode {
 				break
 			}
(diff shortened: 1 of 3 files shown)


pkg/devspace/config/loader/patch/operation.go
```diff
@@ -35,38 +34,66 @@ func (op *Operation) Perform(doc *yaml.Node) error {
...
-	if len(matches) == 0 {
-		if op.Op == opAdd {
-			matches, err = getParents(doc, op.Path)
-			if err != nil {
-				return fmt.Errorf("could not add using path: %s", op.Path)
+	if len(matches) == 0 && op.Op != opAdd {
+		return fmt.Errorf("%s operation does not apply: doc is missing path: %s", op.Op, op.Path)
+	}
+
+	var f func(parent *yaml.Node, match *yaml.Node)
```
FabianKramm, line 41: Please rename this to a more meaningful variable name and make sure to add a comment why this is needed

pkg/devspace/config/loader/patch/operation.go
```diff
@@ -35,38 +34,66 @@ func (op *Operation) Perform(doc *yaml.Node) error {
...
+
+	switch op.Op {
+	case opAdd:
+		f = op.add
+
+		var pathMatches bool
```
FabianKramm, line 47: Remove that variable and use `len(originalMatches) > 0` below

pkg/devspace/config/loader/patch/operation.go
```diff
@@ -35,38 +34,66 @@ func (op *Operation) Perform(doc *yaml.Node) error {
...
+		matches, err = getParents(doc, op.Path)
+		if err != nil {
+			return fmt.Errorf("could not add using path: %s", op.Path)
+		}
+
+		if len(matches) > 0 && pathMatches {
+			if matches[0].Kind == yaml.SequenceNode {
+				matches = originalMatches
+				break
+			} else {
```
FabianKramm, line 68: else not needed as we break in the if statement already

FabianKramm (changes requested):
@carlmontanari thanks for the PR! Please take a look at my comments below

FabianKramm (approved):
@carlmontanari thanks! LGTM!

PR #2215, opened 2022-08-10
fix: uploading empty files

FabianKramm:
/kind bugfix

Fixed an issue where DevSpace would always reupload empty files on initial sync.


sgreene570:
🚢 
Thank you!

PR #2216, opened 2022-08-11
added support for kubeconfig flag in devspace

### LINKED ISSUE #2203: Add support for --kubeconfig flag in devspace
We can pass --kubeconfig flag to kubectl, can we add the same support for devspace. This way user can type in something like this
`devspace <pipeline stage> --kubeconfig=<path to kubeconfig file> -n <namespace>`

instead of typing below commands
`devspace use context <cn>`
`devspace use namespace <ns>`

No

`devspace <pipeline stage> --kubeconfig=<path to kubeconfig file> -n <namespace>`

`KUBECONFIG=<path to kubeconfig file> devspace <pipeline stage>  -n <namespace>`

**Additional context**  
While it's easy to pass kubeconfig file or context using env variable and use context commands, it would really help in adding support for the kubeconfig flag as it would in sync with kubectl command and also one can use this when invoking devspace programmatically

/kind feature

tukobadnyanoba:
/kind feature

resolves #2203 

Fixed an issue where DevSpace ...will have --kubeconfig flag on root level which will allow user to change kube configuration on the fly.

### DIFF
--- a/cmd/flags/flags.go
+++ b/cmd/flags/flags.go
@@ -7,20 +7,19 @@ import (
 
 // GlobalFlags is the flags that contains the global flags
 type GlobalFlags struct {
-	Silent bool
-	NoWarn bool
-	Debug  bool
-
+	Silent                   bool
+	NoWarn                   bool
+	Debug                    bool
+	DisableProfileActivation bool
+	SwitchContext            bool
+	InactivityTimeout        int
+	KubeConfig               string
 	OverrideName             string
 	Namespace                string
 	KubeContext              string
-	Profiles                 []string
-	DisableProfileActivation bool
-	SwitchContext            bool
 	ConfigPath               string
+	Profiles                 []string
 	Vars                     []string
-
-	InactivityTimeout int
 
 	Flags *flag.FlagSet
 }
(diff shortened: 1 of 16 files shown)


cmd/flags/flags.go
```diff
@@ -55,6 +54,7 @@ func SetGlobalFlags(flags *flag.FlagSet) *GlobalFlags {
 	flags.StringVarP(&globalFlags.Namespace, "namespace", "n", "", "The kubernetes namespace to use")
 	flags.StringVar(&globalFlags.KubeContext, "kube-context", "", "The kubernetes context to use")
 	flags.StringSliceVar(&globalFlags.Vars, "var", []string{}, "Variables to override during execution (e.g. --var=MYVAR=MYVALUE)")
+	flags.StringVarP(&globalFlags.KubeConfig, "kubeconfig", "k", "", "kubeconfig path")
```
FabianKramm, line 57: Please don't create a shorthand flag, and use the description `The kubeconfig path to use`

FabianKramm (changes requested):
@tukobadnyanoba thanks for the PR! Please see my comment below

FabianKramm (approved):

Issue #2237, opened 2022-08-23
pipeline flags cannot be used if variable with input is used


FabianKramm:
**What happened?**  
Pipeline flags are not parsed correctly if variables with input are used in the config, e.g.:
```
version: v2beta1
name: flags
vars:
  MY_VAR:
    default: test

pipelines:
  dev:
    flags:
    - name: test
      type: bool
    run: |-
      echo "$(get_flag test)"
```

Running `devspace dev --test` will result in the error `flag --test not found`

PR #2238, opened 2022-08-23
fix: allow variable input with pipeline flags #2237

lizardruss:
### Changes
-  [x] use empty variable values when resolving flags


FabianKramm:
Closed in favor of #2240

PR #2242, opened 2022-08-25
fix: supply helm registry for oci login

CiiDyR:
/kind bugfix

Fixes v6.0.0-beta.0 issue where helm registry login for OCI charts did not supply target registry.

Fixed missing `registry` argument for `helm registry login`

I'm not a Go developer, so have mercy.


FabianKramm (approved):
@CiiDyR thanks for the PR! LGTM!

PR #2244, opened 2022-08-25
fix: use chart directory to detect chart changes

lizardruss:
/kind bugfix

When using a helm chart with no name, the current directory is used to detect chart changes. If the current directory contains the `.devspace` folder, it will always detect changes and redeploy the application. This issue presented when using a git repository chart source in a dependency.

Fixed an issue where DevSpace would always re-deploy charts imported using git.

### DIFF
--- a/pkg/devspace/deploy/deployer/helm/deploy.go
+++ b/pkg/devspace/deploy/deployer/helm/deploy.go
@@ -59,7 +59,10 @@ func (d *DeployConfig) Deploy(ctx devspacecontext.Context, forceDeploy bool) (bo
 		chartPath = ctx.ResolvePath(chartPath)
 
 		// Check if the chart directory has changed
-		hash, err = hashpkg.Directory(chartPath)
+		hash, err = hashpkg.DirectoryExcludes(chartPath, []string{
+			".git/",
+			".devspace/",
+		}, true)
 		if err != nil {
 			return false, errors.Errorf("Error hashing chart directory: %v", err)
 		}
--- a/pkg/devspace/deploy/deployer/helm/deploy.go
+++ b/pkg/devspace/deploy/deployer/helm/deploy.go
@@ -2,10 +2,11 @@ package helm
 
 import (
 	"fmt"
-	"github.com/loft-sh/devspace/pkg/devspace/config/versions"
 	"io"
 	"os"
 	"path/filepath"
+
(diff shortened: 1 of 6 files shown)


pkg/devspace/helm/v3/client.go
```diff
@@ -56,12 +65,11 @@ func (c *client) InstallChart(ctx devspacecontext.Context, releaseName string, r
 	// Chart settings
 	chartName := ""
 	if helmConfig.Chart.Source != nil {
-		chartName, err = dependencyutil.DownloadDependency(ctx.Context(), ctx.WorkingDir(), helmConfig.Chart.Source, ctx.Log())
+		chartName, err = c.DownloadChart(ctx, helmConfig)
```
lizardruss, line 68: This results in two downloads when there are changes. I'm not sure what side effects of modifying the helm config name after the first DownloadChart call would have, so opting for less changes.

pkg/devspace/deploy/deployer/helm/deploy.go
```diff
@@ -44,6 +45,14 @@ func (d *DeployConfig) Deploy(ctx devspacecontext.Context, forceDeploy bool) (bo
 		releaseNamespace = d.DeploymentConfig.Namespace
 	}
 
+	if d.DeploymentConfig.Helm.Chart.Source != nil {
```
lizardruss, line 48: This adds overhead to each to deploy, but hopefully better than redeploying each time.

pkg/devspace/deploy/deployer/helm/deploy.go
```diff
@@ -44,6 +45,14 @@ func (d *DeployConfig) Deploy(ctx devspacecontext.Context, forceDeploy bool) (bo
...
+		downloadPath, err := d.Helm.DownloadChart(ctx, d.DeploymentConfig.Helm)
+		if err != nil {
+			return false, errors.Wrap(err, "download chart")
+		}
+		chartPath = downloadPath
+	}
```
lizardruss, line 54: This adds overhead to each to deploy, but hopefully better than redeploying each time.

pkg/devspace/helm/v3/client.go
lizardruss, line 68: This results in two downloads when there are changes. I'm not sure what side effects of modifying the helm config name after the first DownloadChart call would have, so opting for less changes.

FabianKramm (approved):

PR #2250, opened 2022-08-26
fix: correctly compare semver according to spec 2.0.0

CiiDyR:
/kind bugfix

According to the latest stable semver spec, devspace is currently not correctly comparing semver versions.
We noticed this when a constraint we had set (`>= 6.0.0-beta.4` did not allow the usage of `6.0.1-beta.0`). According to Item 11 of the spec (https://semver.org/spec/v2.0.0.html#spec-item-11), this should have worked.
I tracked this down to a constraint in the `prereleaseCheck` function, that only allowed pre-release versions of the same base segments to be correctly compared. Removing this check still passes all tests but now correctly follows the spec in that regard.
I also removed a duplicate test from the suite.

Fixed an issue where DevSpace incorrectly compared semantic versions. 


FabianKramm (approved):
@CiiDyR thanks for the PR! LGTM!

PR #2297, opened 2022-09-10
Add DEVSPACE_PROFILES variable and is_in command

alexissellier:
/kind enhancement

Add builtin `DEVSPACE_PROFILES` variable containing all enabled profiles

Add `DEVSPACE_PROFILES` variable

For now it is still a draft because the variable is an array and is "converted" into a list of string with extra `[` and `]`. Need to figure out how to resolve that


FabianKramm (approved):
@alexissellier thanks for the PR! LGTM!

PR #2319, opened 2022-09-17
Resolve variables when updating them after profiles are merged

### LINKED ISSUE #2262: vars in a profile not overriding the default vars
**What happened?**

If I try to override a `var` in a profile using the `merge` option, it does not work.

I would expect any `var` defined in a profile to override the default `var` section.

My devspace.yaml:

```
version: v2beta1
name: bug
vars:
  VAR_1: 'value 1'

pipelines:
  build:
    run: |-
      echo $VAR_1

profiles:
- name: asdf
  merge:
    vars:
      VAR_1: 'value 2'
```

Then run `devspace print -p asdf`. You will see that `VAR_1` is set to `value 1`, whereas I would expect it to be set to `value 2`. This worked as expected in the beta version of v6, but stopped working when we upgraded to version 6.0.0 and still does not work on version 6.0.1.

**Local Environment:**

* DevSpace Version: devspace version 6.0.1
* Operating System: linux | mac (seeing on both linux and mac)
* ARCH of the OS: AMD64 | ARM64
  **Kubernetes Cluster:**  N/A
* Cloud Provider: N/A
* Kubernetes Version: N/A

/kind bug

alexissellier:
/kind bugfix

resolves #2262

Fixed an issue where DevSpace variables where not resolved after profile activation that would override an existing variable

### DIFF
--- a/pkg/devspace/config/loader/imports.go
+++ b/pkg/devspace/config/loader/imports.go
@@ -3,6 +3,9 @@ package loader
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
 	"github.com/loft-sh/devspace/pkg/devspace/config/loader/variable"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/util"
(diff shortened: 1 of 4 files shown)


alexissellier:
I will try to work on this before Wednesday

lizardruss (commented):
This looks pretty good to me, but I would like to see a test case added to prevent reverting this change as DevSpace evolves.

alexandradragodan:
Hey, folks 👋 

Thanks for your contribution so far! Are you still working on this PR? 

Please let us know if there's anything we can support with. We'd be happy to!

PR #2326, opened 2022-09-20
Local registry

lizardruss:
/kind feature

Adds a local image registry when user lacks push permissions to the remote registry.

**TODO**
- [x] Detect push permissions
- [x] Deploy registry
- [x] Wait for Registry to be ready
- [x] Start port forwarding
- [x] Rewrite image tag for deployed registry
- [x] Build / Push image
- [x] Update runtime variables
- [x] Update config substitution
- [x] Update `ShouldRebuild()` to account for local registry usage
- [x] Acceptance Tests:
  - [x] get_image output
  - [x] `${runtime.images.my-image.image}` && `${runtime.images.my-image}`
  - [x] `image(name)`
  - [x] kubectl deployment rewritten
  - [x] helm deployment rewritten
  - [x] dev.*.imageSelector (and all the other different runtime rewritten config sections)
  - [x] dev.*.devImage
- [x] Move local registry configuration to top level (avoid duplicating for multiple images)
- [x] Add progress indicator when transferring images
- [x] `devspace init` doesn't produce config for local-registry (even in comments).
- [x] `devspace build` on minikube with default config(no config for local-registry) spins the registry pod doesn't push the image to it and `devspace build` command errors out. (Note: updated to skip push for `devspace build` and minikube, as well as skip local registry for minikube)
- [x] `devspace cleanup local-registry` command asks question (sure to cleanup) even when there is no registry pod running.
- [x] logs need formatting


pkg/devspace/build/registry/local_registry.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return errors.Wrap(err, "wait for node port")
+	}
+
+	// Save registry host for rewriting images
+	r.host = fmt.Sprintf("localhost:%d", r.servicePort.NodePort)
+
+	// Start port forwarding
+	tctx, t := ctx.WithNewTomb()
+	pfctx := tctx.WithContext(context.Background())
```
FabianKramm, line 75: context.Background is not a good idea here. Also it seems we never stop port-forwarding to the registry as long as DevSpace runs, is that a good idea?

pkg/devspace/build/registry/local_registry.go
```diff
@@ -0,0 +1,216 @@
...
+	<-t.NotifyGo(func() error {
```
FabianKramm, line 76: How do we catch errors here that happen during port forwarding?

pkg/devspace/build/registry/local_registry.go
```diff
@@ -0,0 +1,216 @@
...
+		return "", err
+	}
+
+	tag.Registry = registry
+
+	return tag.Repository.Name(), nil
+}
+
+func (r *LocalRegistry) ensureNamespace(ctx devspacecontext.Context) error {
+	_, err := ctx.KubeClient().KubeClient().CoreV1().Namespaces().Create(
```
FabianKramm, line 109: Better to do either an apply here or GET and then CREATE

pkg/devspace/config/versions/latest/schema.go
```diff
@@ -582,6 +585,39 @@ type CustomConfigCommand struct {
 	OperatingSystem string `yaml:"os,omitempty" json:"os,omitempty"`
 }
 
+// LocalRegistryConfig holds the configuration of the local image registry
+type LocalRegistryConfig struct {
+	// Disable disables the local registry when true. True by default
```
FabianKramm, line 590: So the local registry is disabled by default?

FabianKramm (changes requested):

pkg/devspace/config/versions/latest/schema.go
lizardruss: Oops. Will update that comment.

pkg/devspace/build/registry/local_registry.go
lizardruss: I’ll change the context.

We can shut down the port forwarding once all images are pushed.

It was useful to leave it open for checking the contents of the registry, however users likely won’t use it that way.

pkg/devspace/build/registry/local_registry.go
lizardruss: 👍

pkg/devspace/config/versions/latest/schema.go
lizardruss: converted to this schema remove ambiguity
```
localRegistry:
  enabled: *bool
```

PR #2327, opened 2022-09-20
docs: replace chmod/mv with install command

tradel:
/kind documentation

Replace `chmod`/`mv` commands in install docs with single `install -c` command.

Simplified install docs.

Requested by @FabianKramm


FabianKramm (approved):
@tradel thanks for the PR! LGTM!

PR #2330, opened 2022-09-21
feat: fallback to ssh if https doesn't work and vice versa

### LINKED ISSUE #2323: Fallback to ssh if git https doesn't work
Currently you can only define a git dependency either via https:// or git@ (ssh). Would be great if DevSpace would fallback to try https in the ssh case and ssh in the https case.

So the logic would be for `https`:

* Try [https://domain.com/path](https://domain.com/path), if that doesn't work, use git@domain:path

For ssh the logic would be the reverse

pratikjagrut:
/kind feature

resolves #2323

Added a new feature where devspace will use ssh if https fails and vice versa while pulling dependency.


lizardruss (approved):
LGTM!

PR #2345, opened 2022-09-26
feat: load DEVSPACE_ENV_FILE file when specified in a dependency

### LINKED ISSUE #2328: Support env files in dependencies
`.env` file from dependencies should be taken into account

alexissellier:
/kind feature

resolves #2328

Allow loading variables from dotenv file in dependencies

### DIFF
--- a/pkg/devspace/dependency/resolver.go
+++ b/pkg/devspace/dependency/resolver.go
@@ -12,6 +12,7 @@ import (
 	"github.com/loft-sh/devspace/pkg/devspace/dependency/graph"
 	"github.com/loft-sh/devspace/pkg/util/kubeconfig"
 
+	"github.com/joho/godotenv"
 	"github.com/loft-sh/devspace/pkg/devspace/config"
 	"github.com/loft-sh/devspace/pkg/devspace/config/loader"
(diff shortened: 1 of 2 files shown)


pkg/devspace/dependency/resolver.go
```diff
@@ -245,6 +246,16 @@ func (r *resolver) resolveDependency(ctx devspacecontext.Context, dependencyConf
 		return nil, err
 	}
 
+	if val, ok := dConfigWrapper.Variables()["DEVSPACE_ENV_FILE"]; ok {
```
FabianKramm, line 249: Not sure if that is a good way of doing this. The problem here is that the config is already resolved here and any variable that would get changed through the `DEVSPACE_ENV_FILE` wouldn't be reflected in the resolved config. While this probably fixes it for some use cases, I don't think we should go with just that as the variables need to be available before actually parsing the config.

FabianKramm (changes requested):

pkg/devspace/dependency/resolver.go
alexissellier: Yes, it needs to be re-worked.

PR #2355, opened 2022-09-30
fix: fallback to ssh if https fails and vice versa

### LINKED ISSUE #2323: Fallback to ssh if git https doesn't work

pratikjagrut:
/kind bugfix

resolves #2323 

Fixed an issue where DevSpace hangs till git credentials are not provided.

### DIFF
--- a/pkg/util/git/git.go
+++ b/pkg/util/git/git.go
@@ -2,10 +2,11 @@ package git
 
 import (
 	"context"
-	"github.com/loft-sh/loft-util/pkg/command"
-	"mvdan.cc/sh/v3/expand"
 	"os"
 	"strings"
+
+	"github.com/loft-sh/loft-util/pkg/command"
+	"mvdan.cc/sh/v3/expand"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


pkg/util/git/git.go
```diff
@@ -64,7 +65,14 @@ func (gr *GitCLIRepository) Clone(ctx context.Context, options CloneOptions) err
 		}
 
 		args = append(args, options.Args...)
-		out, err := command.CombinedOutput(ctx, gr.LocalPath, expand.ListEnviron(os.Environ()...), "git", args...)
+		os.Setenv("GIT_SSH_COMMAND", "ssh -oBatchMode=yes")
```
lizardruss, line 68: Is this to prevent prompting for a password? If not, a comment with explanation might be helpful.

lizardruss (commented):
Nice work!

I had some questions that might prompt changes, but otherwise it looks good.

pkg/util/git/git.go
```diff
@@ -64,7 +65,14 @@ func (gr *GitCLIRepository) Clone(ctx context.Context, options CloneOptions) err
...
+		gitEnv := []string{
+			"GIT_TERMINAL_PROMPT=0",
+			"GIT_SSH_COMMAND='ssh -oBatchMode=yes'",
+		}
+		gitEnv = append(gitEnv, os.Environ()...)
+		out, err := command.CombinedOutput(ctx, gr.LocalPath, expand.ListEnviron(gitEnv...), "git", args...)
+		os.Unsetenv("GIT_SSH_COMMAND")
```
lizardruss, line 75: Is `os.Setenv()` and `os.Unsetenv()` required, or is passing `gitEnv` when running the command sufficient?

pkg/util/git/git.go
pratikjagrut: Yes, this is for preventing prompts and preventing ssh from asking for adding unknown hosts to known host list. I'll add comments there.

pkg/util/git/git.go
pratikjagrut: It's required. `GIT_SSH_COMMAND ` will be used by the underlying ssh command.

pkg/util/git/git.go
pratikjagrut: Done.

pkg/util/git/git.go
pratikjagrut: Yes, it is, I seem to misunderstand it.

FabianKramm (approved):

PR #2390, opened 2022-10-21
Skip local registry use for vclusters on local kubernetes clusters

lizardruss:
/kind bugfix

- Resolves an issue where the local registry was used for vclusters on Minikube, Docker Desktop, and KinD clusters
- Refactored a few locations that were checking for Minikube that would have failed for vcluster & minikube
- Allow local registry to be used with Minikube when explicitly enabled

Fixed an issue where DevSpace local registry was incorrectly used for virtual clusters on local Kubernetes clusters

Previously the local registry could never be used with minikube. These changes allow the user to opt in to the local registry with Minikube, and makes it consistent with the other local kubernetes options (KinD & Docker Desktop). There isn't a practical reason to do this, but using `localRegistry.enabled=true` in the config, and not having the local registry used was unexpected to some during testing.

There is also a [KinD issue ](https://github.com/kubernetes-sigs/kind/issues/2960)where `--force-build` creates a new tag, but `kind load docker-image` fails to re-tag the image in the KinD cluster, leading to an `ImagePullBackoff`. I've verified this issue exists in DevSpace version before the local registry feature was added. Using the local registry works as a workaround for this issue.

### DIFF
--- a/pkg/devspace/build/build.go
+++ b/pkg/devspace/build/build.go
@@ -11,7 +11,6 @@ import (
 	"github.com/loft-sh/devspace/pkg/devspace/build/types"
 	"github.com/loft-sh/devspace/pkg/devspace/config/constants"
 	devspacecontext "github.com/loft-sh/devspace/pkg/devspace/context"
-	"github.com/loft-sh/devspace/pkg/devspace/kubectl"
 	"github.com/loft-sh/devspace/pkg/util/stringutil"
 
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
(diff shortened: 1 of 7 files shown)


pkg/devspace/build/build.go
```diff
@@ -106,10 +101,7 @@ func (c *controller) Build(ctx devspacecontext.Context, images []string, options
 
 			// Determine whether the local registry is required / enabled
 			isLocalReqistryRequired := !registry.HasPushPermission(imageConf)
-			useMinikubeDocker := registry.UseMinikubeDocker(ctx, imageConf)
-			if useMinikubeDocker {
-				ctx.Log().Warnf("Using Minikube for image %s, skipping local registry", imageConf.Image)
```
lizardruss, line 111: We can bring back this edge case if desired, but it was somewhat unexpected to have `localRegistry.enabled = true` and it still not be used because of minikube.

This did require using a docker client configured for the minikube environment to work, see later comments.

pkg/devspace/build/registry/util.go
```diff
@@ -67,13 +69,13 @@ func IsImageAvailableRemotely(ctx context.Context, imageName string) (bool, erro
...
 
-func CopyImageToRemote(ctx context.Context, imageName string, writer io.Writer) error {
+func CopyImageToRemote(ctx context.Context, client dockerclient.Client, imageName string, writer io.Writer) error {
 	ref, err := name.ParseReference(imageName)
 	if err != nil {
 		return err
 	}
 
-	image, err := daemon.Image(ref, daemon.WithContext(ctx))
+	image, err := daemon.Image(ref, daemon.WithContext(ctx), daemon.WithClient(client.DockerAPIClient()))
```
lizardruss, line 78: This update allows the local registry push to work with either the local docker daemon, or the Minikube docker daemon.

pkg/devspace/docker/client.go
```diff
@@ -107,7 +109,7 @@ func newDockerClientFromEnvironment() (Client, error) {
 }
 
 func newDockerClientFromMinikube(ctx context.Context, currentKubeContext string) (Client, error) {
-	if currentKubeContext != "minikube" {
```
lizardruss, line 110: This would also miss vcluster + Minikube contexts

lizardruss (commented):
@FabianKramm added some additional context to these changes. Let me know if you'd like anything changed.

FabianKramm (approved):

PR #2396, opened 2022-10-26
docs: add profiles partial for dependency

alexissellier:

/kind documentation

Documentation is missing a section about the profile parameter for dependency section


FabianKramm (approved):
@alexissellier thanks for the PR, LGTM!

PR #2400, opened 2022-11-03
Pass on devspace vars to plugin env

mahendrabagul:
/kind enhancement

resolves #

Pass all config vars to a DevSpace plugin command call, so they are available in the DevSpace plugin's environment

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -294,6 +294,9 @@ Additional run commands:
 	rootCmd.AddCommand(NewPurgeCmd(f, globalFlags, rawConfig))
 
 	// Add plugin commands
+	if rawConfig != nil && rawConfig.OriginalRawConfig != nil {
+		plugin.AddDevspaceVarsToPluginEnv(rawConfig.OriginalRawConfig["vars"])
+	}
 	plugin.AddPluginCommands(rootCmd, plugins, "")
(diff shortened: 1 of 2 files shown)


pkg/devspace/plugin/plugin.go
```diff
@@ -346,6 +351,25 @@ func Decode(encoded string) ([]byte, error) {
 	return encoding.DecodeString(encoded)
 }
 
+func AddDevspaceVarsToPluginEnv(vars interface{}) {
+	devspaceVarsOnce.Do(func() {
+		devspaceVarsLock.Lock()
```
FabianKramm, line 356: The lock is not needed here

cmd/root.go
```diff
@@ -294,6 +294,7 @@ Additional run commands:
 	rootCmd.AddCommand(NewPurgeCmd(f, globalFlags, rawConfig))
 
 	// Add plugin commands
+	plugin.AddDevspaceVarsToPluginEnv(rawConfig.OriginalRawConfig["vars"])
```
FabianKramm, line 297: What happens if rawConfig or rawConfig.OriginalRawConfig is nil?

FabianKramm (changes requested):

FabianKramm (approved):

PR #2401, opened 2022-11-03
Add EnsureNamespace call to create a namespace

### LINKED ISSUE #2399: devspace does not create namespace for dependency deployment
**What happened?**
I declared a dependency in my `devspace.yaml` file with the namespace config option set. The namespace does not exist when starting the cluster. The deployment fails because the namespace isn't created.

Devspace should create this namespace automatically (see #605).

My devspace.yaml:

```
version: v2beta1
name: xxx-api

pipelines:
   dev:
    run: |-
      run_dependencies --all      
      create_deployments --all    
      start_dev app                

deployments:
  xxx-api:
    helm:
      chart:
        name: xxx-api
        repo: https://xxx.xxx.net/
        username: ${HELM_USERNAME}
        password: ${HELM_PASSWORD}
      valuesFiles: ["dev/values.yaml"]

dev:
  app:
    devImage: ghcr.io/lusu007/devspace-containers/typescript:18-alpine
    imageSelector: xxx.xxx.xxx:5005/xxx-api:nightly
    sync:
      - path: ./
        uploadExcludeFile: .dockerignore
    terminal:
      command: NPM_TOKEN=${NPM_TOKEN} ./devspace_start.sh
    ssh:
      enabled: true
    proxyCommands:
      - command: devspace
      - command: kubectl
      - command: helm
      - command: git
    ports:
      - port: "9229"
      - port: "3000"
    open:
      - url: http://localhost:3000
vars:
  HELM_PASSWORD:
    password: true
  NPM_TOKEN:
    question: "Please enter your NPM token:"
    password: true

dependencies:
  xxx-api:
    git: https://xxx.xxx/xx/xx/xxxx
    branch: dev
    pipeline: dev
    namespace: xxx-api
```

**Local Environment:**

* DevSpace Version: 6.1.1
* Operating System: Linux Ubuntu (WSL)
* ARCH of the OS: AMD64
  **Kubernetes Cluster:**
* Cloud Provider: kind
* Kubernetes Version:  v1.25.3

mahendrabagul:
/kind bugfix

resolves #2399

Add --create-namespace arg to helm install command to create a namespace if not exists

### DIFF
--- a/e2e/framework/helper.go
+++ b/e2e/framework/helper.go
@@ -63,6 +63,21 @@ func ExpectEmpty(actual interface{}, explain ...interface{}) {
 	gomega.ExpectWithOffset(1, actual).To(gomega.BeEmpty(), explain...)
 }
 
+func ExpectNamespaceExistence(namespace string) {
+	kubeClient, err := kube.NewKubeHelper()
+	ExpectNoErrorWithOffset(1, err)
+
+	err = wait.PollImmediate(time.Second, time.Minute*2, func() (done bool, err error) {
+		ns, err := kubeClient.GetNamespace(namespace)
+		if err != nil {
+			return false, nil
+		}
+
+		return ns.Name == namespace, nil
+	})
+	ExpectNoErrorWithOffset(1, err)
+}
+
 func ExpectRemoteFileContents(imageSelector string, namespace string, filePath string, contents string) {
 	kubeClient, err := kube.NewKubeHelper()
 	ExpectNoErrorWithOffset(1, err)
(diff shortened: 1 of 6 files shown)


pkg/devspace/helm/v3/client.go
```diff
@@ -76,6 +76,7 @@ func (c *client) InstallChart(ctx devspacecontext.Context, releaseName string, r
 		var chartRepo string
 		chartName, chartRepo = generic.ChartNameAndRepo(helmConfig)
 		args = append(args, chartName)
+		args = append(args, "--create-namespace")
```
FabianKramm, line 79: Problem is we cannot do that as `--create-namespace` needs namespace permissions which in spaces are not given. Also this would only solve the case for helm install and not kubectl deployments

FabianKramm (changes requested):

meyfa:
@FabianKramm May I ask why this was closed? It seems like a great addition and would simplify our setup a lot.

PR #2424, opened 2022-11-15
feat: add global flag for running dep sequentially

### LINKED ISSUE #2380: Deploy with `--max-concurrent-builds` or `--build-sequential` does not apply to concurrent dependency builds
**What happened?**
I have a monorepo with API gateway. The gateway project depends on multiple service projects. Each project, including gateway, has at least one image to build.

When I run either `devspace deploy --build-sequential` or `devspace deploy --max-concurrent-builds 1` , devspace kicks off builds for mulitiple dependency projects at once, resulting in heavy parallelism, high memory usage by docker and recurring build failures.

I expected both `devspace deploy --build-sequential` and `devspace deploy --max-concurrent-builds 1` to kick off 1 build at a time. I would like developers to choose concurrency with `devspace deploy --max-concurrent-builds X`,  to control CPU load and memory usage on their machine.

I have prepared a reproduction repo, just follow the steps from the readme.

[https://github.com/gustaff-weldon/devspace-build-concurrency](https://github.com/gustaff-weldon/devspace-build-concurrency)

**Local Environment:**

* DevSpace Version: 6.1.1
* Operating System: mac
* ARCH of the OS: x86_64

**Kubernetes Cluster:**

* local Docker Desktop cluster, but builds happen outside of it

/kind bug

pratikjagrut:
/kind feature

resolves #2380 

Added global flag `--sequential-dependencies` which deploys dependencies sequentially.

### DIFF
--- a/pkg/devspace/pipeline/pipeline.go
+++ b/pkg/devspace/pipeline/pipeline.go
@@ -224,10 +224,13 @@ func (p *pipeline) StartNewDependencies(ctx devspacecontext.Context, dependencie
 
 	// Start sequentially
 	if options.Sequential {
+		ctx.Log().Debug("Deploying dependencies sequentially")
 		for _, dependency := range deployDependencies {
 			err := p.startNewDependency(ctx, dependency, options)
 			if err != nil {
 				return errors.Wrapf(err, "run dependency %s", dependency.Name())
+			} else {
+				ctx.Log().Debugf("Dependency '%s' deployed", dependency.Name())
 			}
 		}
 
--- a/cmd/run_pipeline.go
+++ b/cmd/run_pipeline.go
@@ -3,6 +3,9 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
+	"os"
+
 	"github.com/loft-sh/devspace/cmd/flags"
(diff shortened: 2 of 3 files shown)


pkg/devspace/pipeline/pipeline.go
```diff
@@ -224,10 +224,13 @@ func (p *pipeline) StartNewDependencies(ctx devspacecontext.Context, dependencie
 
 	// Start sequentially
 	if options.Sequential {
+		ctx.Log().Info("Deploying dependencies sequentially")
```
lizardruss, line 227: Can we switch both log statements to `Debug()`/`Debugf()`?

lizardruss (changes requested):
Very small change requested, but otherwise LGTM!

pkg/devspace/pipeline/pipeline.go
pratikjagrut: Okay.

pratikjagrut: Done

FabianKramm (changes requested):

FabianKramm (approved):

PR #2425, opened 2022-11-15
fix: port forward to local registry with multiple dependencies

### LINKED ISSUE #2411: Local registry issues with dependencies
**What happened?**
When using local registry in a DevSpace project with dependencies, only some of the images can be pushed.

All images to be built and pushed.

Using the [microservices example](https://github.com/loft-sh/devspace/tree/main/examples/microservices), the second image to build will attempt to start a second local registry and fail

**Local Environment:**

* DevSpace Version: [v6.2.0-alpha.0](https://github.com/loft-sh/devspace/releases/tag/v6.2.0-alpha.0)

Each dependency build tries to start a local registry with the same name, and then attempt to port-forward on the same port.

**Workaround:**
Configure the local registry for dependents to have different names. This will cause them to have separate ports to forward on.

node/devspace.yaml

```
localRegistry:
  name: node-registry
```

php/devspace.yaml

```
localRegistry:
  name: php-registry
```

lizardruss:
/kind bugfix

resolves #2411
Fixes ENG-603

Fixed an issue where DevSpace could not reuse a local registry if required by multiple dependencies

This is a basic fix, where the registry is matched by name. If dependencies use different registry configurations, then the last one wins.

### DIFF
--- a/pkg/devspace/build/build.go
+++ b/pkg/devspace/build/build.go
@@ -87,7 +87,6 @@ func (c *controller) Build(ctx devspacecontext.Context, images []string, options
 	}
 
 	// Determine if we need to use the local registry to build any images.
-	var localRegistry *registry.LocalRegistry
 	kubeClient := ctx.KubeClient()
 	builders := map[string]builder.Interface{}
 	tags := map[string][]string{}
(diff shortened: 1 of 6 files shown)


carlmontanari:
just tested w/ the same process that created this issue for us in the first place and it is now working as desired! 🔥

PR #2432, opened 2022-11-17
fix: Imported variables are not used in other variables

### LINKED ISSUE #2408: Imported variables cannot be used in other variables
Imported variables cannot be used in other variables.

Example repository: [https://github.com/hendrikheil/devspace-import-issues](https://github.com/hendrikheil/devspace-import-issues)

pratikjagrut:
/kind bugfix

resolves #2408 

Fixed an issue where DevSpace was not using imported variables in defining another variable in main devspace.yaml

### DIFF
--- a/e2e/tests/imports/testdata/conditional/devspace.yaml
+++ b/e2e/tests/imports/testdata/conditional/devspace.yaml
@@ -3,13 +3,20 @@ name: base
 
 imports:
   - path: ./${IMPORT1_PATH}
+  - path: ./import.yaml
 
 vars:
   IMPORT1_PATH:
     source: env
     default: import1.yaml
+  TODAYS_MESSAGE:
+    source: command
+    command: echo ${NEW_MESSAGE}
+  HELLO_WORLD: ""
 
 pipelines:
   deploy:
     run: |-
       echo ${IMPORT1} > import1.txt
+      echo ${TODAYS_MESSAGE} > message.txt
+      echo ${HELLO_WORLD} > hello.txt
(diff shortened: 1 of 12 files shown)


pkg/devspace/config/loader/variable/resolver.go
```diff
@@ -386,7 +386,13 @@ func (r *resolver) resolve(ctx context.Context, name string, definition *latest.
...
 
 	// set variable so that we don't ask again
-	r.memoryCache[name] = value
+	if v, ok := value.(string); ok {
+		if v != "" {
+			r.memoryCache[name] = value
+		}
+	} else {
+		r.memoryCache[name] = value
+	}
```
lizardruss, line 395: This might make it so the importer cannot set a variable to `""`.

e2e/tests/imports/testdata/conditional/devspace.yaml
```diff
@@ -3,13 +3,18 @@ name: base
...
   - path: ./${IMPORT1_PATH}
+  - path: ./import.yaml
 
 vars:
   IMPORT1_PATH:
     source: env
     default: import1.yaml
+  TODAYS_MESSAGE:
+    source: command
+    command: echo ${NEW_MESSAGE}
```
lizardruss, line 14: Can we add a case for when both have a var with the same name, where the importer sets the value to `""`?

I suspect that the imported value would be used over the local value, which may be unexpected.

lizardruss (changes requested):
There could be unexpected behavior with the current implementation. I'll ask for a second opinion as well.

FabianKramm (approved):

PR #2455, opened 2022-12-05
Perhaps add a note that the sample values break a deploy

gregmsanderson:
Either add a note like this _or_ replace the shown kustomizeArgs values with ones that _do_ work? I tried the manifest as shown and they break the deploy since that is not a valid kustomize command (at least not with the _latest_ version of it, perhaps they used to work):

`create_deployments: error deploying app: error executing 'kustomize build ./k8s/base --timeout 30s --grace-period 30': Error: unknown flag: --timeout`

/kind documentation

Text edit to clarify

Updated docs


FabianKramm (approved):

FabianKramm:
@gregmsanderson thanks a lot for the PR! LGTM!

PR #2500, opened 2023-01-16
Feature: Add support for InlineManifest blocks in Kubectl deployments

89luca89:
/kind feature

This should make possible to use inline manifests in `kubectl` blocks, for example:

```yaml
version: v2beta1
name: inline-manifest

vars:
  IMAGE: nginx:1.14.2

deployments:
  my-deployment:
    kubectl:
      inlineManifest: |-
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: nginx-deployment
          labels:
            app: nginx
        spec:
          replicas: 3
          selector:
            matchLabels:
              app: nginx
          template:
            metadata:
              labels:
                app: nginx
            spec:
              containers:
              - name: nginx
                image: ${IMAGE}
                ports:
                - containerPort: 80
```

- [x] Need to update tests with the new use case
- [x] I specified that it is not possible to have both `manifests` and `inlineManifest` at the same time, but we might want to revise this decision?
- [x] For now the example in `examples/inlineManifest` is not working, need to work on that

### DIFF
--- a/pkg/devspace/config/versions/validate.go
+++ b/pkg/devspace/config/versions/validate.go
@@ -256,8 +256,11 @@ func validateDeployments(config *latest.Config) error {
 		if deployConfig.Helm == nil && deployConfig.Kubectl == nil {
 			return errors.Errorf("Please specify either helm or kubectl as deployment type in deployment %s", deployConfig.Name)
 		}
-		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil {
-			return errors.Errorf("deployments[%s].kubectl.manifests is required", index)
+		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil && deployConfig.Kubectl.InlineManifest == "" {
+			return errors.Errorf("deployments[%s].kubectl.manifests or deployments[%s].kubectl.InlineManifest is required", index, index)
+		}
+		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests != nil && deployConfig.Kubectl.InlineManifest != "" {
+			return errors.Errorf("deployments[%s].kubectl.manifests and deployments[%s].kubectl.inlineManifest cannot be used together", index, index)
 		}
 		if deployConfig.Kubectl != nil && deployConfig.Helm != nil {
 			return errors.Errorf("deployments[%s].kubectl and deployments[%s].helm cannot be used together", index, index)
--- a/pkg/devspace/deploy/deployer/kubectl/kubectl.go
+++ b/pkg/devspace/deploy/deployer/kubectl/kubectl.go
@@ -4,13 +4,15 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/loft-sh/devspace/pkg/devspace/config/constants"
 	"io"
-	"mvdan.cc/sh/v3/expand"
 	"os"
 	"strings"
 
+	"github.com/loft-sh/devspace/pkg/devspace/config/constants"
+	"mvdan.cc/sh/v3/expand"
+
 	"github.com/loft-sh/devspace/pkg/devspace/config/loader/variable/legacy"
+	"github.com/loft-sh/devspace/pkg/devspace/config/loader/variable/runtime"
 	"github.com/loft-sh/devspace/pkg/devspace/config/remotecache"
 	devspacecontext "github.com/loft-sh/devspace/pkg/devspace/context"
 	"github.com/loft-sh/devspace/pkg/devspace/context/values"
@@ -30,14 +32,17 @@ import (
 	"github.com/loft-sh/devspace/pkg/util/hash"
 )
 
+var Cachemanifest = "./.devspace/manifest-cache.yaml"
+
 // DeployConfig holds the necessary information for kubectl deployment
 type DeployConfig struct {
-	Name        string
-	CmdPath     string
-	Context     string
-	Namespace   string
-	IsInCluster bool
-	Manifests   []string
+	Name           string
+	CmdPath        string
+	Context        string
+	Namespace      string
+	IsInCluster    bool
+	InlineManifest string
+	Manifests      []string
 
 	DeploymentConfig *latest.DeploymentConfig
 }
@@ -46,7 +51,7 @@ type DeployConfig struct {
 func New(ctx devspacecontext.Context, deployConfig *latest.DeploymentConfig) (deployer.Interface, error) {
 	if deployConfig.Kubectl == nil {
 		return nil, errors.New("error creating kubectl deploy config: kubectl is nil")
-	} else if deployConfig.Kubectl.Manifests == nil {
+	} else if deployConfig.Kubectl.Manifests == nil && deployConfig.Kubectl.InlineManifest == "" {
 		return nil, errors.New("no manifests defined for kubectl deploy")
 	}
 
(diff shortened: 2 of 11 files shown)


pkg/devspace/config/versions/validate.go
```diff
@@ -256,8 +256,11 @@ func validateDeployments(config *latest.Config) error {
 		if deployConfig.Helm == nil && deployConfig.Kubectl == nil {
 			return errors.Errorf("Please specify either helm or kubectl as deployment type in deployment %s", deployConfig.Name)
 		}
-		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil {
-			return errors.Errorf("deployments[%s].kubectl.manifests is required", index)
+		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil && deployConfig.Kubectl.InlineManifest == "" {
+			return errors.Errorf("deployments[%s].kubectl.manifests or deployments[%s].kubectl.InlineManifest is required", index)
+		}
+		if deployConfig.Kubectl.Manifests != nil && deployConfig.Kubectl.InlineManifest != "" {
```
FabianKramm, line 262: This will run into a panic if `deployConfig.Kubectl` is nil

pkg/devspace/deploy/deployer/kubectl/kubectl.go
```diff
@@ -46,7 +49,7 @@ type DeployConfig struct {
 func New(ctx devspacecontext.Context, deployConfig *latest.DeploymentConfig) (deployer.Interface, error) {
 	if deployConfig.Kubectl == nil {
 		return nil, errors.New("error creating kubectl deploy config: kubectl is nil")
-	} else if deployConfig.Kubectl.Manifests == nil {
+	} else if deployConfig.Kubectl.Manifests == nil && deployConfig.Kubectl.InlineManifest == " " {
```
FabianKramm, line 52: Use `""` instead of `" "`

pkg/devspace/deploy/deployer/kubectl/kubectl.go
```diff
@@ -74,6 +77,16 @@ func New(ctx devspacecontext.Context, deployConfig *latest.DeploymentConfig) (de
 		manifests = append(manifests, manifest)
 	}
 
+	if deployConfig.Kubectl.InlineManifest != "" {
+		// save created manifest to file
+		// fmt.Println(deployConfig.Kubectl.InlineManifest)
```
FabianKramm, line 82: Please remove the debug output

pkg/devspace/deploy/deployer/kubectl/kubectl.go
```diff
@@ -74,6 +77,16 @@ func New(ctx devspacecontext.Context, deployConfig *latest.DeploymentConfig) (de
...
+		err := os.WriteFile(Cachemanifest, []byte(deployConfig.Kubectl.InlineManifest), 0644)
```
FabianKramm, line 83: This can cause problems with multiple deployments using the same file at the same time, please try to use an approach without a temporary file and pass in the bytes directly

FabianKramm (changes requested):

pkg/devspace/config/versions/validate.go
89luca89: Done

pkg/devspace/deploy/deployer/kubectl/kubectl.go
89luca89: Done

pkg/devspace/config/versions/validate.go
89luca89: Done

pkg/devspace/deploy/deployer/kubectl/kubectl.go
89luca89: Yes about this, it was one point where I wanted to think more

This way, we're recycling the current `Deploy()` without changes
Right now we're just using the `DeployConfig.Manifests` 

One approach could be expanding the `DeployConfig` struct, with the inline manifest data, and do a specific apply for them

This would also unlock the possibility of having both manifests (paths) and inline manifests

What do you think?

FabianKramm: I'm not sure if having `inlineManifests` and `manifests` together is a good idea, as the order would be undefined and its easy to just define 2 deployments which would deploy those

pkg/devspace/config/versions/validate.go
```diff
@@ -256,8 +256,11 @@ func validateDeployments(config *latest.Config) error {
 		if deployConfig.Helm == nil && deployConfig.Kubectl == nil {
 			return errors.Errorf("Please specify either helm or kubectl as deployment type in deployment %s", deployConfig.Name)
 		}
-		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil {
-			return errors.Errorf("deployments[%s].kubectl.manifests is required", index)
+		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests == nil && deployConfig.Kubectl.InlineManifest == "" {
+			return errors.Errorf("deployments[%s].kubectl.manifests or deployments[%s].kubectl.InlineManifest is required", index, index)
+		}
+		if deployConfig.Kubectl != nil && deployConfig.Kubectl.Manifests != nil && deployConfig.Kubectl.InlineManifest != "" {
+			return errors.Errorf("deployments[%s].kubectl and deployments[%s].inlineManifest cannot be used together", index, index)
```
lizardruss, line 263: Should this message be the following?
```golang
"deployments[%s].kubectl.manifests and deployments[%s].kubectl.inlineManifest cannot be used together"
```

lizardruss (changes requested):
Left a comment about an error message that might need correcting.

Adding an E2E test similar to this [kubectl test](https://github.com/loft-sh/devspace/blob/main/e2e/tests/deploy/deploy.go#L275-L313) might be good as well.

Really nice work!

pkg/devspace/config/versions/validate.go
89luca89: You're right :sweat_smile: Fixed it now :+1: 

Will also work on the e2e :+1:

FabianKramm (approved):

FabianKramm:
As per discussion:
1. We need to cover the following cases (1,2 & 3) for `kubectl.inlineManifests`:
```
version: v2beta1
name: main

images:
  your-image:
    image: my-registry:5000/my-repo/my-image

deployments:
  deploy-via-inline-kubectl:
    kubectl:
      inlineManifest: |-
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: test
        spec:
          replicas: 1
          selector:
            matchLabels:
              run: test
          template:
            metadata:
              labels:
                run: ${DEVSPACE_VAR} # <- CASE 1 (Regular Devspace VAR)
            spec:
              containers:
                - name: ${runtime.images.your-image} # <- CASE 2 (Runtime Variable)
                  image: my-registry:5000/my-repo/my-image # <- CASE 3 (Image Name)
```
2. To cover these cases do the following:
  * Case 1: Nothing left todo, this will be covered automatically
  * Case 2: To accomplish this, before you work with the value `Kubectl.InlineManifests`, you do:
```
	resolvedInlineManifests, err := runtime.NewRuntimeResolver(ctx.WorkingDir(), false).FillRuntimeVariablesAsString(ctx.Context(), deployConfig.InlineManifests, ctx.Config(), ctx.Dependencies())
```
  * Case 3: Reuse the existing functionality in getReplacedManifests. Adjust the function to accept also a string

FabianKramm (approved):

PR #2510, opened 2023-01-24
docs: add back runtime variables page from 5.x. Fix ENG-878

89luca89:
/kind documentation

Adding back the Runtime Variables page in the docs. It is present DevSpace v5 docs but not in v6 anymore

Fixes ENG-878


lizardruss (approved):
LGTM

PR #2512, opened 2023-01-25
fix: add pull secret config for images during v5->v6 schema upgrade

lizardruss:
/kind bugfix

Fixes ENG-922

Fixed an issue where DevSpace no longer creates image pull secrets for all images for v1beta11 schemas


89luca89:
LGTM

PR #2513, opened 2023-01-26
feat: add SkipPushKindControlPlane option in image configuration

alexissellier:
/kind feature

Add an option called `skipPushOnKindControlPlane` on to only target worker node when pushing docker images using kind binary

Add an option to prevent images to be push on kind control plane node,

### DIFF
--- a/cmd/run_pipeline.go
+++ b/cmd/run_pipeline.go
@@ -39,11 +39,12 @@ import (
 type RunPipelineCmd struct {
 	*flags.GlobalFlags
 
-	Tags                    []string
-	Render                  bool
-	Pipeline                string
-	SkipPush                bool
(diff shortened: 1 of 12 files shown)


FabianKramm:
Hey @alexissellier, thanks for the PR! Seems like the e2e tests are failing, would you mind taking a look?

FabianKramm:
@alexissellier this seems like a very specific use case, I'm not sure if we should add another global pipeline flag just for that as we already have way too many flags. Isn't there another way how we could achieve this without any flag?

alexissellier:
I can only add a configuration parameter if it suit you

alexissellier:
@FabianKramm I've removed all reference to a flag and only keep a config option

PR #2543, opened 2023-02-09
feat: add a parameter to disable helm dependency update

### LINKED ISSUE #2533: Allow a way to ignore helm depdencies
Yes, we currently have a helm chart that we disable dependencies for via conditions because we don't use them, but since Devspace updates dependencies anyway it will download them. We've found that a server serving the dependencies is flaky and is causing our devspace setup to also be flaky.

Allow a way for users to specify if dependencies should be updated on install/upgrade.

Don't update dependencies and instead just run an upgrade/install.

**Additional context**
The helm chart is Temporal.

alexissellier:
/kind feature

resolves #2533

Add a parameter to disable helm dependency update

### DIFF
--- a/devspace-schema.json
+++ b/devspace-schema.json
@@ -744,6 +744,10 @@
           },
           "type": "array",
           "description": "TemplateArgs are additional arguments to pass to `helm template`"
+        },
+        "disableDependencyUpdate" : {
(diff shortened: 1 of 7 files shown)


lizardruss (approved):
Thank you! LGTM.

PR #2558, opened 2023-02-14
fix: continue even if a dev config has no Ports when running list ports

### LINKED ISSUE #2557: list ports if one entry in devspace
**What happened?**  

When running `devspace list ports`, it displays `No ports are forwarded` even if there are some configured

List all the ports mentioned in the config 

**Local Environment:**  
- DevSpace Version: 6.2.5
- Operating System: linux
- ARCH of the OS: AMD64

alexissellier:
/kind bugfix

resolves #2557

Fixed an issue where DevSpace doe not list forwarded ports when one entry in the dev section has none configured

### DIFF
--- a/cmd/list/ports.go
+++ b/cmd/list/ports.go
@@ -2,6 +2,7 @@ package list
 
 import (
 	"context"
+
 	"github.com/loft-sh/devspace/cmd/flags"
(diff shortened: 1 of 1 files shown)


lizardruss (approved):
LGTM

PR #2559, opened 2023-02-15
feat: add a json output for list port command

alexissellier:
/kind enhancement

Add a json output to `devspace list ports` to align with other command

Add a json output to devspace list ports

### DIFF
--- a/cmd/list/ports.go
+++ b/cmd/list/ports.go
@@ -2,6 +2,8 @@ package list
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


alexissellier:
Maybe we could define a custom tag like `table: ImageSelector` and  have a custom to display the information as a table instead of log.PrintTable

FabianKramm (approved):
@alexissellier thanks for the PR! LGTM!

PR #2561, opened 2023-02-16
feat: add patches to kubectl deployments

### LINKED ISSUE #1678: Add patches/resources via kustomize deployment
Would be nice to support patches and referencing additional resources directly in kustomize deployment spec.

Temporary kustomization.yaml file can be created as an overlay, this file can absorb orginal kustomize settings plus additional patches and resources from devspace.yaml to prevent inconsistency of original kustomization.

```
deployments:
- name: my-deployment
  kubectl:
    kustomize: true
    manifests:
    - .
    patches:
    - path: ingress_patch.json
      target:
        group: networking.k8s.io
        version: v1beta1
        kind: Ingress
        name: my-ingress
    - target:
        kind: Deployment
        name: deploy
      patch: |-
      - op: replace
        path: /spec/template/spec/containers/0/image
        value: nginx:latest
    resources:
    - https://github.com/lukasmrtvy/kustomize//raw.yaml?ref=master
```

```
hooks:
- command: kustomize
  args:
    - edit
    - add
    - patch
    - --group=networking.k8s.io
    - --version=v1
    - --kind=Ingress
    - --name=${INGRESS_OBJECT_NAME}
    - --patch=[{...}]
  when:
    before:
      deployments: all
```

There is a chance that users can commit modified version of kustomization.yaml ( even with hook which will remove patch ), beacuse `kustomize edit add patch` command actualy modifies kustomization.yaml file itself.

**Additional context**

Thanks

/kind feature

89luca89:
/kind enhancement
/kind feature

resolves #1678
Closes ENG-249

This PR adds support for patches also in kubectl deployments.
It supports both normal and inlineManifests

Works like the profile patches, so `add`, `remove`, `replace` operations are supported
Wildcards are supported

Example yaml:

```yaml
version: v2beta1
name: nginx-k8s

deployments:
  example:
    kubectl:
      inlineManifest: |-
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: nginx-deployment
        spec:
          selector:
            matchLabels:
              app: nginx
          replicas: 4
          template:
            metadata:
              labels:
                app: nginx
            spec:
              containers:
                - name: nginx
                  image: nginx:1.14.2
                  ports:
                    - containerPort: 80
      patches:
        - target:
            apiVersion: apps/v1 # Optional
            kind: Deployment    # Optional
            name: nginx-deployment # Required
          op: replace
          path: spec.template.spec.containers[0].image
          value: nginx:1.23.3
```

### DIFF
--- a/pkg/devspace/deploy/deployer/kubectl/kubectl.go
+++ b/pkg/devspace/deploy/deployer/kubectl/kubectl.go
@@ -359,7 +359,7 @@ func (d *DeployConfig) applyDeployPatches(ctx devspacecontext.Context, resource 
 			Path: patch.OpPath(patch.TransformPath(kubepatch.Path)),
 		}
 
-		if kubepatch.Name != resource.GetName() {
+		if kubepatch.Target.Name != resource.GetName() {
 			continue
 		}
 
@@ -384,7 +384,7 @@ func (d *DeployConfig) applyDeployPatches(ctx devspacecontext.Context, resource 
 		}
 
 		// TODO Maybe log here that we're indeed applying a patch?
-		ctx.Log().Debugf("applying patch: %s.%s", kubepatch.Name, kubepatch.Path)
+		ctx.Log().Debugf("applying patch: %s.%s", kubepatch.Target.Name, kubepatch.Path)
 		patches = append(patches, newPatch)
 	}
 
--- a/pkg/devspace/deploy/deployer/kubectl/kubectl.go
+++ b/pkg/devspace/deploy/deployer/kubectl/kubectl.go
@@ -3,7 +3,6 @@ package kubectl
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io"
 	"os"
 	"strings"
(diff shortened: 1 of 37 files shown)


pkg/devspace/deploy/deployer/kubectl/kubectl.go
```diff
@@ -338,3 +345,60 @@ func (d *DeployConfig) isKustomizeInstalled(ctx context.Context, dir, path strin
...
+	out, err := yaml.Marshal(resource)
+	if err != nil {
+		return resource, err
+	}
+
+	patches := patch.Patch{}
+	for idx, kubepatch := range d.DeploymentConfig.Kubectl.Patches {
+		newPatch := patch.Operation{
+			Op:   patch.Op(kubepatch.Operation),
+			Path: patch.OpPath(kubepatch.Path),
```
lizardruss, line 359: Would we we want to support the older style of paths by adding the transform step that profiles take? It could be confusing for users to have the format accepted in one area, but not the other.

Similar to this [line](https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/config/loader/profile.go#L246).

lizardruss (changes requested):
One question about supporting the same paths as profiles, but otherwise awesome work!

FabianKramm:
@89luca89 thanks a lot for the PR! LGTM!

lizardruss (approved):
LGTM!

PR #2568, opened 2023-02-21
fix: set force-build flag to false by default

### LINKED ISSUE #2567: build option force-rebuild is always true
**What happened?**

When building an image, the `--force-rebuild` flag is always true leading to image rebuilding even if the docker context did not change or even if `rebuildStrategy: ignoreContextChanges` is specify

I expect the flag to be set to ``` false ``` by default and its value controlled by \`\`--force-rebuild\`

Build any image you want using the `build_image` pipeline function

alexissellier:
/kind bugfix

resolves #2567

Fixed an issue where DevSpace sets force-build to false by default

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -12,7 +12,6 @@ func NewBuildCmd(f factory.Factory, globalFlags *flags.GlobalFlags, rawConfig *R
 	cmd := &RunPipelineCmd{
 		GlobalFlags:             globalFlags,
 		Pipeline:                "build",
-		ForceBuild:              true,
 		SkipPushLocalKubernetes: true,
 	}
 


cmd/build.go
```diff
@@ -12,7 +12,6 @@ func NewBuildCmd(f factory.Factory, globalFlags *flags.GlobalFlags, rawConfig *R
 	cmd := &RunPipelineCmd{
 		GlobalFlags:             globalFlags,
 		Pipeline:                "build",
-		ForceBuild:              true,
```
lizardruss, line 15: Hello! This is only used as the default value, and can be changed to false by passing `--force-build=false`. The code that allows the flag value to be used is [this call](https://github.com/devspace-sh/devspace/blob/main/cmd/build.go#L36), which [configures cobra to use the passed flag](https://github.com/devspace-sh/devspace/blob/main/cmd/run_pipeline.go#L75).

PR #2574, opened 2023-02-27
fix: add implicit patch during dev to unset readOnlyRootFilesystem when sync is enabled

### LINKED ISSUE #2572: Add implicit patch during dev to unset readOnlyRootFilesystem

Our base Helm chart has `securityContext.readOnlyRootFilesystem` set to `true` on its container to improve production security. During development with `devspace dev`, this causes the following error message:

```
start_dev: start sync: error executing tar: tar: can't remove old file devspacehelper: Read-only file system
: command terminated with exit code 1
fatal exit status 1
```

The docs describe a number of implicit patches applied by DevSpace to dev containers: [https://www.devspace.sh/docs/configuration/dev/modifications/patches#implicit-patches](https://www.devspace.sh/docs/configuration/dev/modifications/patches#implicit-patches)

I propose adding the following patch:

```
- op: remove
  path: spec.containers[].securityContext.readOnlyRootFilesystem
```

In my opinion, this makes sense since a read-only filesystem could never work with DevSpace, so this should be unset by default.

We are currently applying this patch manually, which works fine, but clutters devspace.yaml. For reference, this is part of our devspace.yaml:

```
dev:
  app:
    patches:
      - op: remove
        path: spec.containers.name=${DEVSPACE_NAME}.securityContext.readOnlyRootFilesystem
```

**Additional context**

I'm using DevSpace version 6.3.0.

89luca89:
Sync may not work if the container SecurityContext enables ReadOnlyRootFilesystem.
So let's do an Implicit Patch to remove this, and leave user's `devspace.yaml` less cluttered

/kind bugfix

resolves #2572 
Closes ENG-1055

### DIFF
--- a/docs/pages/cli/devspace_list_ports.md
+++ b/docs/pages/cli/devspace_list_ports.md
@@ -25,7 +25,8 @@ Lists the port forwarding configurations
 ## Flags
 
 ```
-  -h, --help   help for ports
+  -h, --help            help for ports
+  -o, --output string   The output format of the command. Can be either empty or json
 ```
 
 
--- a/docs/pages/configuration/dev/modifications/patches.mdx
+++ b/docs/pages/configuration/dev/modifications/patches.mdx
@@ -62,7 +62,8 @@ DevSpace adds certain modifications and patches in-memory if certain conditions 
   <tr>
 <td>
 
-`terminal` enabled
+`terminal` enabled or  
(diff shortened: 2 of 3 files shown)


FabianKramm (approved):

FabianKramm:
@89luca89 thanks a lot for the PR! LGTM!

PR #2582, opened 2023-03-02
updateImageTags actually defaults to true

opticoder:
/kind bugfix
/kind documentation

fix wrong updateImageTags default value mention in docs

fix updateImageTags default value mention in docs

### DIFF
--- a/docs/pages/configuration/_partials/v2beta1/deployments/updateImageTags.mdx
+++ b/docs/pages/configuration/_partials/v2beta1/deployments/updateImageTags.mdx
@@ -2,7 +2,7 @@
 <details className="config-field" data-expandable="false" open>
 <summary>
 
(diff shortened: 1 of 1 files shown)


FabianKramm:
@opticoder thanks for the PR! It actually defaults to `true`, so in the docs its actually correct

opticoder:
@FabianKramm Thanks for the quick reply! But doesn't it mean then here https://www.devspace.sh/docs/configuration/deployments/#deployments-updateImageTags default value is not correct? And if yes, should I change the PR or create a new one?

FabianKramm:
@opticoder oh yes! You are correct its actually wrong in the other place!

opticoder:
updated the fix for default to true

lizardruss:
Closing in favor of #2714, since there were fixes to the documentation generation code needed to avoid overwriting this change.

PR #2584, opened 2023-03-03
feat: support local builds with docker/podman for local registry

89luca89:
/kind enhancement
/kind feature

Closes ENG-1027

This commit will reimplement local builds using docker/podman when using local registry.
To preserve the current default behaviour to use buildkit, this is implemented using a new configuration `localbuild: true/false`

```yaml
# The local registry will not use buildkit on the cluster, but your docker/podman daemon
# and then push it to the local registry in the cluster
localRegistry:
  localbuild: true
```

When `localbuild:true` devspace 
- will deploy only the `registry` container (skipping buildkitd) in the cluster
- use docker/podman locally to build the image
- portforward the registry port in localhost
- copy the local image to registry

PR includes also updated docs and e2e for the use case


pkg/devspace/config/versions/latest/schema.go
```diff
@@ -596,6 +596,9 @@ type LocalRegistryConfig struct {
 	// When `false` the local registry will never be used.
 	Enabled *bool `yaml:"enabled,omitempty" json:"enabled,omitempty"`
 
+	// LocalBuild enables use of local docker builder instead of building in the cluster
+	LocalBuild *bool `yaml:"localbuild,omitempty" json:"localbuild,omitempty"`
```
FabianKramm, line 600: Why is this a pointer?

FabianKramm (changes requested):
@89luca89 thanks a lot for the PR! Very good! 1 small comment besides that looks good to me!

pkg/devspace/config/versions/latest/schema.go
89luca89: Fixed :+1:

FabianKramm (approved):

lizardruss (approved):
LGTM! Thank you!

PR #2344, opened 2022-09-26
[WIP] feat: add tanka integration to Devspace

xvzf:
/kind feature
resolves https://github.com/loft-sh/devspace/issues/2269

## Purpose
This PR picks up on #2325 and introduces Tanka environment support for Devspaces. We internally use Tanka and would like to share a common code base between all environments, including local development stacks.

## Todos
- [x] Extend the schema definition with a `tanka` deployment configuration
- [ ] Fix the schema of deployments.tanka
- [ ] Add a Tanka deployer under pkg/devspace/deploy/deployer/tanka
- [ ]  Check if we add an option to run jb update before using tk
- [ ] Make a PR to https://github.com/loft-sh/loft-util to be able to download tanka binary
- [ ] Add e2e tests

Closes #2325

PR #2594, opened 2023-03-09
Combine stdout and stderr when getting version from requried commands

jordiclariana:
/kind enhancement

Some required commands output their version on stderr instead of stdout. Devspace is not handling stderr when running `command version` so it is not possible to assess that the command is installed with its right version. This change changes the behavior so stdout and stderr are combined and used the same to look for the version string/

Combine stdout and stderr for required commands when getting their version

This can potentially break compatibility since some programs can mix version data between stdout and stderr output. I assume this is an edge case that should not be common at all but is still possible.

### DIFF
--- a/pkg/devspace/config/loader/loader.go
+++ b/pkg/devspace/config/loader/loader.go
@@ -269,7 +269,7 @@ func (l *configLoader) ensureRequires(ctx context.Context, config *latest.Config
 			return errors.Wrapf(err, "parsing require.commands[%d].version", index)
 		}
 
-		out, err := command.Output(ctx, filepath.Dir(l.absConfigPath), expand.ListEnviron(os.Environ()...), c.Name, versionArgs...)
+		out, err := command.CombinedOutput(ctx, filepath.Dir(l.absConfigPath), expand.ListEnviron(os.Environ()...), c.Name, versionArgs...)
 		if err != nil {
 			aggregatedErrors = append(aggregatedErrors, fmt.Errorf("cannot run command '%s' (%v), however it is required by the config. Please make sure you have correctly installed '%s' with version %s", c.Name, err, c.Name, c.Version))
 			continue


lizardruss (commented):
Do you have an example command where this change is needed to support it?

In general combining the two outputs may not be ideal. It's possible to add a small script to invoke the command and redirect stderr to stdout. Would that be an acceptable workaround?

jordiclariana:
Hi @lizardruss, here're some examples I'm working on:

_Full disclosure: I'm also contributing to [this PR](https://github.com/devspace-sh/devspace/pull/2344) which adds Tanka support to devspace._

* [Tanka](https://github.com/grafana/tanka/releases): `tk --version`
* [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler/releases): `jb --version`

I tend to agree with you that combining stdout and stderr is not in general a good thing to do, I've written my opinion on that in my initial comment. But I think that in this case is the less dangerous one. We are using a well-known particular switch (`version`, `--version`, `-V`, ...) to get the version, and stdout and stderr mixing should not be something that might collide with each other in this case. Most likely, either there is stdout only or stderr only. That is at least my rationality.

As an alternative, we might add another field under `require.command[]` that specifies which output we want to capture, but I thought that can be overkill for such a task. I can develop it though if you think is better.

jordiclariana:
I missed this in my comment:
> It's possible to add a small script to invoke the command and redirect stderr to stdout

I guess you're referring on having a wrapper to capture the stderr and throw it into stdout. That is possible, although

1. either a wrapper with the name of the original tool needs to created per tool (since we use `require.commands[].name` as the executable to run)
```yaml
require:
  commands:
    - name: tk_wrapper.sh
      versionArgs: ["--version"]
    - name: jb_wrapper.sh
      versionArgs: ["--version"]
```

2. or use the same `name` for all commands with different versionArgs:
```yaml
require:
  commands:
    - name: wrapper.sh
      versionArgs: ["tk", "--version"]
    - name: wrapper.sh
      versionArgs: ["jb", "--version"]
```

On option 1 I think it can be too cumbersome to maintain. Option 2 I have doubts it might work (right now I don't know if name collision is checked). In any case, none of these look to me elegant solutions, although I agree it will maintain compatibility (if there is any non-compatibility on my original approach).

PR #2604, opened 2023-03-21
fix: do ensureNamespace only if the namespace doesn't really exist.

### LINKED ISSUE #2603: Running local image registry should not require permission to patch "namespaces" resource if the namespace already exists
**What happened?**

Running `devspace build` results in the error

```
local-registry: Starting Local Image Registry
build_images: build images: create builder: get or create local registry: ensure namespace: namespaces "toddqu" is forbidden: User "u-gy5a5otqg6" cannot patch resource "namespaces" in API group "" in the namespace "toddqu"
```

The namespace "toddqu" already exists, I expect the registry to be deployed to it.  If it already exists, why is it trying to patch the "namespaces" resource?  I don't have permission for that, and I don't expect it will be granted to me.

I did try this running against a local Rancher Desktop cluster where I do have permissions for that and it worked as expected.

Run `devspace build` with a cluster your context doesn't have permission to patch the resource "namespaces" in.

My devspace.yaml:

```
version: ...
...
```

**Local Environment:**

* DevSpace Version: 6.3.1
* Operating System: mac
* ARCH of the OS: ARM64
  **Kubernetes Cluster:**
* Cloud Provider: other - on-prem Rancher
* Kubernetes Version: 1.23.6

Traced it to I think this function, but didn't see any way to work around it myself.  [https://github.com/devspace-sh/devspace/blob/94f1ea07600ecb085ed76e8d264c4e6d6eb6af94/pkg/devspace/build/localregistry/local_registry.go#L135](https://github.com/devspace-sh/devspace/blob/94f1ea07600ecb085ed76e8d264c4e6d6eb6af94/pkg/devspace/build/localregistry/local_registry.go#L135)

89luca89:
/kind bugfix

resolves #2603 
Closes ENG-1154

Fixed an issue where DevSpace would attempt to create the namespace for localregistry even it the namespace indeed exists.
This came out as a problem in the moment the user wasn't authorized to create namespaces, thus checking before leaping now.

### DIFF
--- a/pkg/devspace/build/localregistry/local_registry.go
+++ b/pkg/devspace/build/localregistry/local_registry.go
@@ -36,7 +36,10 @@ type LocalRegistry struct {
 	servicePort *corev1.ServicePort
 }
 
-func GetOrCreateLocalRegistry(ctx devspacecontext.Context, options Options) (*LocalRegistry, error) {
+func GetOrCreateLocalRegistry(
+	ctx devspacecontext.Context,
+	options Options,
+) (*LocalRegistry, error) {
 	localRegistriesLock.Lock()
 	defer localRegistriesLock.Unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/devspace/build/localregistry/local_registry.go
```diff
@@ -164,6 +167,32 @@ func (r *LocalRegistry) RewriteImageForBuilder(image string) (string, error) {
...
 func (r *LocalRegistry) ensureNamespace(ctx devspacecontext.Context) error {
+	// If localregistry namespace is the same as devspace, we don't have
+	// anything to do.
+	if r.Namespace == ctx.KubeClient().Namespace() {
+		return nil
+	}
+
+	// List all namespaces, this will already return an error in case of
+	// user's permissions problems.
+	namespaces, err := ctx.KubeClient().
```
lizardruss, line 178: Why list and not get?

lizardruss (commented):
May be able to simplify with a different API call, but I could be missing context.

pkg/devspace/build/localregistry/local_registry.go
89luca89: I thought about it, but I wanted to be sure that an `err` from the client was for permission related issues
Because `namespace XXX does not exist` is also an `err` in the get, and I thought pattern-matching err wouldn't be optimal

lizardruss: There is a function that helps with the error detection. [Here’s an example usage](https://github.com/devspace-sh/devspace/blob/b13898a69ec1431f584b4a5e3d4e26c13b3e645b/cmd/cleanup/local_registry.go#L100).

89luca89: Oh Thanks will look into it!

89luca89: Done :) much better this way, thanks!

lizardruss (approved):
LGTM!

PR #2609, opened 2023-03-23
fix: ensure namespaces during dependencies resolution

### LINKED ISSUE #2399: devspace does not create namespace for dependency deployment

89luca89:
/bugfix

resolves #2399 
Closes ENG-1089

During dependencies resolution, let's ensure the dependency's namespace is existing
This also unifies the localregistry ensurenamespace code, in order to simplify things

### DIFF
--- a/e2e/tests/deploy/deploy.go
+++ b/e2e/tests/deploy/deploy.go
@@ -1,10 +1,15 @@
 package deploy
 
 import (
+	"bytes"
 	"context"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/loft-sh/devspace/pkg/util/log"
+	"github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
 
 	"github.com/onsi/ginkgo/v2"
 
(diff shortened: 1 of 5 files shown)


pkg/devspace/pipeline/pipeline.go
```diff
@@ -223,6 +223,10 @@ func (p *pipeline) StartNewDependencies(ctx devspacecontext.Context, dependencie
 			continue
 		}
 
+		err := ensureNamespace(ctx, dependency.DependencyConfig().Namespace)
+		if err != nil {
+			return errors.Wrapf(err, "cannot run dependency %s", dependency.Name())
+		}
```
lizardruss, line 229: What would you think of moving this into `startNewDependency()`? It might be out of place here, and moving it would allow for concurrent and sequential execution, but I'm not sure that matters either way 🤷

lizardruss (changes requested):
Looks good. One possible change suggested to help with sequential vs concurrency execution.

Also adding some e2e tests to prevent regressions would be good. If not too much trouble, it would be nice to have a test for the permission denied scenario so that we know the error is surfaced properly to the user.

pkg/devspace/pipeline/pipeline.go
89luca89: > What would you think of moving this into startNewDependency()?

Right, seems to be the more appropriate position :+1:  thanks!

89luca89:
> Also adding some e2e tests to prevent regressions would be good. If not too much trouble, it would be nice to have a test for the permission denied scenario so that we know the error is surfaced properly to the user.

Yep on it!

meyfa:
@89luca89 It seems like this may have been closed unintentionally when you updated your fork via the GitHub UI? I've seen this happen sometimes... Unfortunately it wipes out the commits on GitHub, but they should still be available locally for you.

89luca89:
Hi @meyfa yea it happened, but I've opened a new one to push the fixes :+1:

PR #2624, opened 2023-04-14
resolve require section from imports

dtomasi:
/kind bugfix

This PR adds a special handling of the `require` section in `loader.ResolveImports`. Instead of overwriting command and plugin constraints, it appends them to the existing ones. This was not handled correctly before, as imported "libraries" overwrote the existing constraints.

It also changes the handling of constraints for `devspace` versios by adding the version constraints form imported files to a already existing constraint, separated by comma. 
Example: `>=6.0, >=6.1`

Fixed an issue where DevSpace did not resolve command and plugin constraints from imports

### DIFF
--- a/pkg/devspace/config/loader/imports.go
+++ b/pkg/devspace/config/loader/imports.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/errors"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @dtomasi 

Thanks a lot for the PR!
Can I ask to add an e2e test for this? So we're sure we don't have regressions in the future

PR #2631, opened 2023-04-24
fix: consider helm release name when showing helm deployment status

### LINKED ISSUE #2626: Listing deployments does not consider releaseName
**What happened?**  
When setting a releaseName for a deployment based on helm charts, as here

```
deployments:
  kafka:
    helm:
      releaseName: ${DEVSPACE_NAME}-kafka
```

and then deploying it via `devspace deploy`, `devspace list deployments` cannot detect the correct status of the deployment.

The reason seems to be that it only uses the deployment name itself to compare against the deployed helm releases.

I'd expect that `devspace list deployments` correctly matches helm releases with deployment configs, e.g. if `devspace deploy` succeeded, `devspace list deplyments` should list all deployments with a `DEPLOYED` status.

```
version: v2beta1
name: quickstart

deployments:
  quickstart:
    helm:
      releaseName: "gotcha"
      values:
        containers:
        - image: loftsh/javascript:latest
```

**Local Environment:**  
- DevSpace Version: 6.3.2
- Operating System: mac, linux
- ARCH of the OS: AMD64
**Kubernetes Cluster:**  
- Cloud Provider: eks
- Kubernetes Version: Client Version: v1.24.2, Kustomize Version: v4.5.4, Server Version: v1.22.17-eks-48e63af

We use the `releaseName` to inject some environment variable. With `devspace v5` this seemed to have worked fine, because the  deployment config was actually a list where we could set the name. Now it is an object and the keys are only allowed to contain alphanumeric characters, it seems.

mschoenlaub:
/kind bugfix

resolves #2626

Fixed an issue where DevSpace would not consider the configured helm release name when showing deployment statuses.

I wonder if some work should be done to avoid the potential of two deployments being configured as having the same helm release name or whether this would be considered a user error which wouldn't happen too often anyway.

### DIFF
--- a/pkg/devspace/deploy/deployer/helm/status.go
+++ b/pkg/devspace/deploy/deployer/helm/status.go
@@ -3,6 +3,7 @@ package helm
 import (
 	"fmt"
 	devspacecontext "github.com/loft-sh/devspace/pkg/devspace/context"
+	"github.com/loft-sh/devspace/pkg/devspace/helm/types"
 
 	"github.com/loft-sh/devspace/pkg/devspace/deploy/deployer"
 	"github.com/loft-sh/devspace/pkg/devspace/helm"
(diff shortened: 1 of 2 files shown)


FabianKramm (approved):
@mschoenlaub thanks a lot for the PR! LGTM!

PR #2662, opened 2023-06-30
Improve docs

lianmakesthings:
/kind documentation

Improves the docs by making some descriptions more specific
- Specify which name will be overridden with the `--override-name` flag
- Mark `name` property not required, but explain what happens when it's not provided


89luca89 (approved):
LGTM

lizardruss (approved):

PR #2666, opened 2023-07-05
fix: add localRegistry to imported sections

### LINKED ISSUE #2665: `localRegistry` property is not "importable"
**What happened?**

When importing a *devspace.yaml* config into another config, it is expected the latter one inherits "all" properties from the parent *devspace.yaml* file. However, this is not happening for the `localRegistry` property.

If a parent *devspace.yaml* file has the `localRegistry` property defined, and if we import it into another *devspace.yaml* we should inherit these changes as well.

Let's say I have the following structure:

```
.
├── devspace.yaml
└── generic
    └── devspace.yaml
```

On the "generic" *devspace.yaml*, I have the following content:

```
version: v2beta1
name: generic

localRegistry:
  enabled: false

vars:
  MY_VAR: MY_VALUE

require:
  commands:
    - name: helm
      version: "> 3.0.0"
```

On the other, I have the following:

```
version: v2beta1
name: service

imports:
  - path: ./generic/devspace.yaml
```

When I do a `devspace print`, I get the following result:

```
❯ devspace print

-------------------

Vars:

               NAME             |               VALUE
  ------------------------------+------------------------------------
    DEVSPACE_CONTEXT            | minikube
    DEVSPACE_EXECUTABLE         | /usr/local/bin/devspace
    DEVSPACE_KUBECTL_EXECUTABLE | kubectl
    DEVSPACE_NAME               | service
    DEVSPACE_NAMESPACE          | default
    DEVSPACE_PROFILE            |
    DEVSPACE_PROFILES           |
    DEVSPACE_RANDOM             | VlHtHE
    DEVSPACE_SPACE              | default
    DEVSPACE_TIMESTAMP          | 1688574458
    DEVSPACE_TMPDIR             | /tmp
    DEVSPACE_USER_HOME          | /home/user
    DEVSPACE_VERSION            | 6.3.2
    MY_VAR                      | MY_VALUE
    devspace.context            | minikube
    devspace.namespace          | default


-------------------

Loaded path: /devspace.yaml

-------------------

version: v2beta1
name: service
require:
    commands:
        - name: helm
          version: '> 3.0.0'
```

Both the `MY_VAR` variable and the `require` property are there, but not the `localRegistry` one.

**Local Environment:**

* DevSpace Version: 6.3.2
* Operating System: linux
* ARCH of the OS: AMD64

lizardruss:
/kind bugfix

resolves #2665
Fixes ENG-1686

Fixed an issue where DevSpace did not import localRegistry configuration

Added a section to the docs to describe import sections that are importable.

### DIFF
--- a/devspace-schema.json
+++ b/devspace-schema.json
@@ -2068,7 +2068,7 @@
     },
     "name": {
       "type": "string",
-      "description": "Name specifies the name of the DevSpace project and uniquely identifies a project.\nDevSpace will not allow multiple active projects with the same name in the same Kubernetes namespace."
+      "description": "Name specifies the name of the DevSpace project and uniquely identifies a project.\nDevSpace will not allow multiple active projects with the same name in the same Kubernetes namespace.\nIf not provided, DevSpace will use the name of the current working directory."
     },
     "imports": {
       "items": {
--- a/docs/pages/configuration/_partials/v2beta1/name.mdx
+++ b/docs/pages/configuration/_partials/v2beta1/name.mdx
@@ -7,6 +7,7 @@
 Name specifies the name of the DevSpace project and uniquely identifies a project.
 DevSpace will not allow multiple active projects with the same name in the same Kubernetes namespace.
 If not provided, DevSpace will use the name of the current working directory.
+
 </summary>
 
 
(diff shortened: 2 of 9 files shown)


FabianKramm (approved):

PR #2668, opened 2023-07-06
docs: clarify where functions can be used

lianmakesthings:
/kind documentation

Improves docs by clarifying that custom `functions` can only be called in the `pipelines` context


FabianKramm (approved):
@lianmakesthings thanks a lot! LGTM!

PR #2684, opened 2023-08-01
fix: run_watch --excludes with ./ not excluded

lizardruss:
Fixed an issue where DevSpace would not exclude run_watch --excluded containing ./

/kind bugfix

Fixes ENG-1820

Fixed an issue where DevSpace would not exclude run_watch --excluded containing ./


pascalbreuninger (approved):
LGTM

89luca89 (approved):
LGTM!

PR #2695, opened 2023-08-16
fix(kubectl): remove the use of the --short flag in kubectl version 

### LINKED ISSUE #2694: Unable to deploy/dev after new kubectl version (1.27 => 1.28)
**What happened?**  

A new version of kubectl was released (1.28). In this version, for the `kubectl version` command, the `--short` flag was depreciated. This flag runs here: https://github.com/devspace-sh/devspace/blob/v6.3.2/pkg/devspace/deploy/deployer/kubectl/builder.go#L82

When I deploy/dev I get:

```
create_deployments: error deploying my-thing: error executing 'kubectl version --client --short': error: unknown flag: --short
```

This only happens when trying to deploy a Kubernetes manifest directly, if deploying using helm charts there is no problem.

The service would deploy/dev

```bash
# ensure you are using a homebrew version of kubectl
which kubectl
# should return /opt/homebrew/bin/kubectl
brew upgrade kubectl
kubectl version
# should return:
# Client Version: v1.28.0
# Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
# Server Version: v1.24.7
# WARNING: version difference between client (1.28) and server (1.24) exceeds the supported minor version skew of +/-1

# then
devspace deploy
```

devspace version 6.3.2
ARM64 Mac

JoshInLisbon:
It was depreciated in kubectl v1.28

/kind bugfix

resolves #2694

Fixed an issue where DevSpace ran the `kubectl version` command with the flag `--short`. This is depreciated in v1.28 of kubectl, so DevSpace would no longer deploy K8s manifests. We use the --output=json flag to standardise the `kubectl version` result across versions. 

Coauthored with @chattsm

### DIFF
--- a/pkg/devspace/deploy/deployer/kubectl/builder.go
+++ b/pkg/devspace/deploy/deployer/kubectl/builder.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
+	"encoding/json"
 
 	jsonyaml "github.com/ghodss/yaml"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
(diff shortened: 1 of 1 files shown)


pkg/devspace/deploy/deployer/kubectl/builder.go
```diff
@@ -78,12 +79,29 @@ func NewKubectlBuilder(path string, config *latest.DeploymentConfig, kubeConfig
...
+    type ClientVersion struct {
+        GitVersion string `json:"gitVersion"`
+    }
+
+    type Data struct {
+        ClientVersion ClientVersion `json:"clientVersion"`
+    }
+
+    var parsedData Data
+    err = json.Unmarshal([]byte(out), &parsedData)
```
lizardruss, line 97: There's a struct available for import from `kubectl`. I've created an alternate PR that uses it and included you as a co-author of the commit.

lizardruss (commented):
Thanks so much for the PR!

I created an [alternate PR](https://github.com/devspace-sh/devspace/pull/2696) based off these changes, and added you as co-authors. It required some vendor changes, but more importantly I wanted to run e2e tests with the new kubectl version.

JoshInLisbon:
Closing because of @lizardruss's PR.

PR #2696, opened 2023-08-16
fix: remove the use of the --short flag in kubectl version

### LINKED ISSUE #2694: Unable to deploy/dev after new kubectl version (1.27 => 1.28)

lizardruss:
/kind bugfix

resolves #2694

Fixed an issue where DevSpace ran the kubectl version command with the flag --short. This is depreciated in v1.28 of kubectl, so DevSpace would no longer deploy K8s manifests. We use the --output=json flag to standardise the kubectl version result across versions.

Fixes ENG-1890


.github/workflows/e2e-tests.yaml
```diff
@@ -21,18 +21,28 @@ env:
...
+      uses: azure/setup-kubectl@v3
+      with:
+        version: ${{ matrix.kubectl_version }}
+
     # Creates KinD with using k8s versions from the matrix above
     - name: Set up kind with K8s version v1.22.4
       uses: engineerd/setup-kind@v0.5.0
       with:
         version: "v0.11.1"
         image: kindest/node:v1.22.4
```
ThomasK33, line 44: We probably should also adjust the node image version to the kubectl version, in which case we should rename the matrix variable name to `kubernetes version` or just `kubernetes`.

```suggestion
        image: kindest/node:${{ matrix.kubectl_version }}
```

ThomasK33 (commented):
Mostly LGTM, minor remark about testing Kubernetes versions in addition to kubectl versions

JoshInLisbon:
Thanks for the PR @lizardruss . 

We were wondering when this might be merged + released in a stable version?

.github/workflows/e2e-tests.yaml
lizardruss: This gets trickier, since KinD doesn't distribute all versions (`v1.27.4` is not available for example), and also not all KinD images are compatible with newer versions. I'll give it a shot, but we may need to manage the versions separately.

ThomasK33 (approved):

lizardruss:
@JoshInLisbon We will aim for a release next week.

JoshInLisbon:
Hi @lizardruss, would it be possible to provide an update on when there might be a new stable release? At the moment I am using a locally produced binary of Devspace. I would like to be able to use a stable release again. This also impacts adoption among other developers.

lizardruss:
@JoshInLisbon DevSpace v6.3.3 has been released and contains this fix.

PR #2699, opened 2023-08-23
docs: Add dark mode

hrittikhere:
/kind documentation

Fixed an issue where DevSpace docs didn't have dark mode :)
In continuation of as they're kind of similar implementation: https://github.com/loft-sh/devpod/pull/607
https://github.com/loft-sh/vcluster/pull/1100


hrittikhere:
Here's the live demo: https://deploy-preview-2699--devspace-docs.netlify.app/docs

I think there's a requirement for Netlify to configure some rules to direct the URI to /docs 👀

FabianKramm (changes requested):
Couple of problems still here (logo background not correct, also generally it looks design-wise off), so we should wait with merging this

hrittikhere:
Do you think the logo looks off design-wise, or are you referring to the color scheme in dark mode? Would you like me to make any changes, or leave it as is till we get back to it again in future?

PR #2710, opened 2023-09-06
Clarify cleanup command

lianmakesthings:
When executing `devspace cleanup images` only the images that have been defined in the `devspace.yaml` are being deleted. This was not explicit in the docs


lizardruss (changes requested):
Thanks Lian! I think to avoid this from being re-generated, you may need to update the descriptions [here](https://github.com/devspace-sh/devspace/blob/main/cmd/cleanup/images.go#L25-L31), and run `gendocs.sh`

PR #2712, opened 2023-09-08
fix: allow SSH to be enabled and disabled via variable

### LINKED ISSUE #2693: Dev Pipeline Keeps SSH-ing
**What happened?**  
`devspace dev` keeps sshing into the pod even when  `ssh.enabled` is set to false.

It should only try to sync the files, without sshing into the pod at the end.

Any example should work, just disable ssh in `devspace.yaml`

My devspace.yaml:
```yaml
version: v2beta1
name: some_name

  dev:
    run: |-
      run_dependencies --all       # 1. Deploy any projects this project needs (see "dependencies")
      ensure_pull_secrets --all    # 2. Ensure pull secrets
      build_images --all -t $(git describe --always)
      create_deployments --all     # 3. Deploy Helm charts and manifests specfied as "deployments"
      start_dev app                # 4. Start dev mode "app" (see "dev" section)

  deploy:
    run: |-
      run_dependencies --all                            # 1. Deploy any projects this project needs (see "dependencies")
      ensure_pull_secrets --all                         # 2. Ensure pull secrets
      build_images --all -t $(git describe --always)    # 3. Build, tag (git commit hash) and push all images (see "images")
      create_deployments --all                          # 4. Deploy Helm charts and manifests specfied as "deployments"

vars:
  USERNAME:
    source: env
    default: dev

images:
  app:
    image: some_image
    dockerfile: path/to/Dockerfile
    context: .

deployments:
  app:
    helm:
      chart:
        name: app/
      values:
        user: ${USERNAME}
        image:
          tag: ${runtime.images.app.tag}

dev:
  app:
    # Search for the container that runs this image
    terminal:
      disableReplace: true
    labelSelector:
      app: app-name
      user: ${USERNAME}
    sync:
      - path: ./
        excludeFile: .gitignore
    ssh:
      enabled: false
    proxyCommands:
      - gitCredentials: true
    ports:
      - port: "3000"

commands:
  migrate-db:
    command: |-
      echo 'This is a cross-platform, shared command that can be used to codify any kind of dev task.'
      echo 'Anyone using this project can invoke it via "devspace run migrate-db"'
```

**Local Environment:**  
- DevSpace Version: `DevSpace version : 6.3.2`
- Operating System: Linux - Fedora 38
- ARCH of the OS: AMD64 
**Kubernetes Cluster:**  
- K3D
- Kubernetes Version: 
```
Client Version: v1.27.4
Kustomize Version: v5.0.1
Server Version: v1.26.1+k3s1
``` 

lizardruss:
/kind bugfix

Fixes #2693
Fixes ENG-1886

Fixed an issue where DevSpace did not allow explicitly disabling SSH when other SSH was given.

This also enables disabling SSH via variables.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/loft-sh/devspace/pkg/util/ptr"
 	"mvdan.cc/sh/v3/expand"
 
 	"github.com/loft-sh/devspace/pkg/devspace/compose"
@@ -759,7 +760,7 @@ func (cmd *InitCmd) addDevConfig(config *latest.Config, imageName, image string,
 	devConfig.DevImage = devImage
 
 	devConfig.SSH = &latest.SSH{
-		Enabled: true,
+		Enabled: ptr.Bool(true),
 	}
 
 	if devConfig.ProxyCommands == nil {
--- a/e2e/tests/ssh/ssh.go
+++ b/e2e/tests/ssh/ssh.go
@@ -2,7 +2,6 @@ package ssh
 
 import (
 	"context"
-	"github.com/onsi/ginkgo/v2"
 	"os"
 	"os/exec"
 	"time"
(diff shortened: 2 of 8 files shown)


lizardruss:
@FabianKramm This is ready for review, but the e2e tests are really flakey. The issue seems to be that the helm deploy fails intermittently, sometimes (but not always) accompanied by a `Downloading helm...` log message, but the runner should already have `helm` installed, and was able to find it on previous test runs it seems. We can always remove the tests if we're confident in the solution. I haven't been able to reproduce the test flakiness locally either.

PR #2716, opened 2023-09-12
fix: remote cache key should consider helm release name

### LINKED ISSUE #2525: Unable to purge helm releases with dynamic releaseName in some cases
**What happened?**
We have a devpsace configuration for a project that deploys different versions into the same namespace. With v6 we can use `releaseName` to deploy these different versions, but when we attempt to clean them up with `devspace purge` the wrong one can be uninstalled and the older ones enter a state where `devspace purge` will not remove them. It seems like when there are multiple versions deployed, the most recently deployed one is what gets purged and then the rest have to be dealt with manually.

Given I have two deployed versions:

```
VERSION="foo" devspace deploy
VERSION="bar" devspace deploy
```

I would expect to purge version `foo` by running the following:

```
VERSION="foo" devspace purge
```

And in the case that I have removed version `bar` first, then I would expect to be able to purge `foo` with a subsequent run of the purge command.

Deploy the same project in the same namespace twice, and make sure the project deploys a helm release. The deployments should use  a different `releaseName` so that there are two helm releases active in the namespace.

Then, use `devpsace purge`. You can attempt to "target" the older helm release by using the same `releaseName` that you created it with, or you can run it without any vars. Either way, you should see the the release that was created second was purged.

Try to run `devpsace purge` again and notice that no helm releases are uninstalled even though the first one is still installed.

My devspace.yaml:

Previously, in v5, we used a variable in the deployment name to prevent conflicts between these versions. Something like this:

```
deployments:
- name: example-app-${VERSION}
```

On v6, we are doing:

```
deployments:
  example-app:
    helm:
      releaseName: example-app-${VERSION}
```

**Local Environment:**

* DevSpace Version: 6.2.5
* Operating System:  linux and mac
* ARCH of the OS: AMD64 and ARM64
  **Kubernetes Cluster:**
* Cloud Provider: aws
* Kubernetes Version: 1.22.0

mschoenlaub:
/kind bugfix

resolves #2525 

Fixed an issue where DevSpace would leave behind orphaned helm releases when setting a helm release name

It doesn't fully fix the use case in #2525, as the originally expected behavior doesn't play well with `devspace` currently.
However, this fix makes `devspace` avoid "orphaned" deployments lingering around.
IMO this is the behavior most people would expect, as the default is to purge all deployments associated with a project.

### DIFF
--- a/pkg/devspace/deploy/deployer/helm/deploy.go
+++ b/pkg/devspace/deploy/deployer/helm/deploy.go
@@ -69,7 +69,7 @@ func (d *DeployConfig) Deploy(ctx devspacecontext.Context, forceDeploy bool) (bo
 	}
 
 	// Ensure deployment config is there
-	deployCache, _ := ctx.Config().RemoteCache().GetDeployment(d.DeploymentConfig.Name)
+	deployCache, _ := ctx.Config().RemoteCache().GetDeployment(releaseName)
 
 	// Check values files for changes
 	helmOverridesHash := ""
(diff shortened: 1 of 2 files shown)


lizardruss (approved):
Thanks for the PR! Looks good to me.

PR #2739, opened 2023-10-18
change: when stopping upstream/downstream, also close grpc connections

### LINKED ISSUE #2738: Leftover GRPC connection, if downstream is gone
**What happened?**  

We're using the devspace module in our code for filesyncing to pods.
However, if we delete the downstream pod, `sync.Stop()` is correctly called, but we're still continuously flooded with grpc logs saying `[core] [Channel #25 SubChannel #26] grpc: addrConn.createTransport failed to connect to {Addr: "", ServerName: "", }. Err: connection error: desc = "transport: failed to write client preface: websocket: close sent"`.

It seems like even though during the `Ping()`, it's recognized that the connection is down and reader/writer are closed, but the connection is left around. Apparently, even canceling the parent context doesn't do the trick here~, probably since `Dial()` will just use `context.Background` (see https://github.com/devspace-sh/devspace/blob/main/helper/util/conn.go#L17).~

I modified the generated [remote_grpc.pb.go ](https://github.com/devspace-sh/devspace/blob/main/helper/remote/remote_grpc.pb.go#L186) file to include a `Drop()` method and call that in the `sync.Stop()` routine - that does the trick:
```Go
func (c *downstreamClient) Drop() error {
	return c.cc.(*grpc.ClientConn).Close()
}
```

I'm sure there's a more elegant way to achieve this though, especially since this is a generated file.

EDIT: This works for me: https://github.com/devspace-sh/devspace/pull/2739

~I've also tried passing around a context via `DialContext`, but that apparently didn't do the trick.~

No flood of logs from GRPC - no dangling connections

iwilltry42:
/kind bugfix

resolves #2738 

Fixed an issue where DevSpace left over orphaned GRPC connections when pods are removed.

This reduces the annoying grpc logs to 2 coming from the ping, just before the connection gets closed.

### DIFF
--- a/helper/server/downstream.go
+++ b/helper/server/downstream.go
@@ -4,9 +4,6 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"context"
-	"github.com/loft-sh/devspace/helper/util/pingtimeout"
-	"github.com/loft-sh/devspace/pkg/util/fsutil"
-	logpkg "github.com/loft-sh/devspace/pkg/util/log"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


FabianKramm (approved):
@iwilltry42 thanks a lot for the PR! LGTM, good catch!

PR #2747, opened 2023-10-26
pass namespace to kubectl build to fix incluster bug

### LINKED ISSUE #2745: Odd namespace behaviour incluster for kubectl deploy
**What happened?**  

I'm trying to deploy a preview using devspace in CI (runners running in kubernetes) in a different namespace than where devspace is running. This is using the kubectl deploy method.

Despite providing the namespace via an arg -> `devspace deploy -n ci-tplkjs-1` devspace ends up deploying the preview in the same namespace as the runner (ci).

There are interesting things to note here and I tried a few things with various level of success.
- the manifests do not have any namespace set
- the logs correctly state that it will use the expected namespace  `info Using namespace 'ci-tplkjs-1'`  but ends up creating in namespace `ci`
- setting `namespace: ci-tplkjs-1` in the deployment definition of devspace doesn't seem to help
- this works properly when running locally with a kubeconfig that defaults to `ci` for the namespace (not incluster)
- forcing a namespace on the manifest directly works 
 - doing so via a patch using devspace works as well	
- forcing a namespace via `applyArgs` and `createArgs` also works

setting namespace via `devspace deploy -n` should be enough to pass down to the manifests when running incluster.

For this you need
- a way to run devspace inside a kubernetes cluster
- enough permission to create a namespace and create resources in that namespace
- a simple devspace deploy configuration using a kubectl deploy method and a manifest that does not have any namespace set

**Local Environment:**  
- DevSpace Version: [use `devspace --version`] v6.3.0
- Operating System: linux
- ARCH of the OS: AMD64
**Kubernetes Cluster:**  
- Cloud Provider: aws
- Kubernetes Version: 1.25

Initially I thought this was the reason https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/kubectl/util/util.go#L31
but according to https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/kubectl/util/util.go#L35-L37 it should only override it if the namespace was empty and it shouldn't be by that point.

I'm down to look deeper into this, I didn't get a chance yet and thought to post here first in case maybe this is intended?

kuuji:
/kind bugfix

resolves #2745

Fixes an issue where DevSpace, when ran `incluster`, would deploy kubectl manifest files in the wrong namespace even when `-n` is provided on the command line.

I'm not setup to run the e2e for this yet but I did test the fix with kubectl (not yet with kustomize). 
Since it's such a simple fix I figured I might as well open the PR to get some opinions while I do more testing.

### DIFF
--- a/pkg/devspace/deploy/deployer/kubectl/builder.go
+++ b/pkg/devspace/deploy/deployer/kubectl/builder.go
@@ -25,7 +25,7 @@ import (
 
 // Builder is the manifest builder interface
 type Builder interface {
-	Build(ctx context.Context, environ expand.Environ, dir, manifest string) ([]*unstructured.Unstructured, error)
+	Build(ctx context.Context, environ expand.Environ, dir, namespace string, manifest string) ([]*unstructured.Unstructured, error)
 }
 
 type kustomizeBuilder struct {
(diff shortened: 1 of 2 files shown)


lizardruss:
@kuuji Thanks for the PR! It's always great to see community contributions!

I approved the e2e test run, so you should see some feedback through the GitHub Actions.

PR #2753, opened 2023-11-07
Devspace defaults to uploading credentials to cluster

### LINKED ISSUE #2729: Devspace defaults to uploading credentials to cluster
**What happened?**  
Running `devspace build` will, by default and without prompting, upload users credentials to the remote cluster.

At a minimum, a prompt requesting permission to propagate user secrets to remote servers.

> `devspace build`
info Using namespace 'xxxxxxx'
info Using kube context 'xxxxxxx'
Ensuring image pull secret for registry: xxxxxxx
Created image pull secret xxxxxxx

My devspace.yaml:
```
version: ...
...
```

**Local Environment:**  
- DevSpace Version: 6.3.2
- Operating System:  mac
- ARCH of the OS: AMD64
**Kubernetes Cluster:**  
- Cloud Provider: google
- Kubernetes Version: v1.25.11-gke.1700

Any organization that uses SSO likely uses local login credentials for services like artifactory, etc... The result being that unless the team using devspace reads about this default behavior, their corporate credentials are now sitting unencrypted in a shared environment without their knowledge.
I can see from the documentation here https://www.devspace.sh/docs/5.x/configuration/pullSecrets/basics that this was a conscious decision. I understand that this was probably to make use of the tool smoother, but IMO is definitely not being handled correctly.
This absolutely should not be the default behavior, and if it is, the user should be prompted before devspace copies credentials to a remote server. Adding a note to the prompt that tells the user how to make the prompt go away for next time (maybe some kind of config) would allow them to quickly transition to that smooth workflow without making security assumptions on their behalf.

lizardruss:
Fixes #2729
Fixes ENG-2185

/kind bugfix

resolves #2729

Fixed an issue where DevSpace creates pull secrets during`devspace build` even when they are not required.

### DIFF
--- a/pkg/devspace/build/build.go
+++ b/pkg/devspace/build/build.go
@@ -1,9 +1,10 @@
 package build
 
 import (
+	"strings"
+
 	dockerclient "github.com/loft-sh/devspace/pkg/devspace/docker"
 	"github.com/loft-sh/devspace/pkg/devspace/pullsecrets"
-	"strings"
 
 	"github.com/loft-sh/devspace/pkg/devspace/build/builder"
 	"github.com/loft-sh/devspace/pkg/devspace/build/types"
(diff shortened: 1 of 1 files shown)


pkg/devspace/build/build.go
```diff
@@ -98,16 +99,28 @@ func (c *controller) Build(ctx devspacecontext.Context, images []string, options
...
+					err = pullsecrets.NewClient().EnsurePullSecret(ctx, dockerClient, pullSecretNamespace, registryURL)
 					if err != nil {
 						ctx.Log().Errorf("error ensuring pull secret for registry %s: %v", registryURL, err)
 					}
 				}
 
-				err = pullsecrets.NewClient().EnsurePullSecret(ctx, dockerClient, ctx.KubeClient().Namespace(), registryURL)
-				if err != nil {
-					ctx.Log().Errorf("error ensuring pull secret for registry %s: %v", registryURL, err)
+				if imageConf.BuildKit != nil && imageConf.BuildKit.InCluster != nil {
```
FabianKramm, line 114: I don't think this is necessary as buildkit is using the local credentials automatically

PR #2761, opened 2023-11-23
docs: move is_dependency to correct function group

jackwilsdon:
/kind documentation

Moves `is_dependency` from the "Global Functions" section of the [built-in functions list](http://localhost:8000/docs/configuration/functions/#built-in-functions) to the "Pipeline-Only Functions" section (`is_dependency` is pipeline-only).

Updated classification of `is_dependency` in documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,3 +47,10 @@ cd ../ # main project directory
 
 go run ./docs/hack/config/schemas/main.go
 ```
+
+### Generate Function Docs
+```bash
+cd ../ # main project directory
+
+go run ./docs/hack/functions/main.go
+```
--- a/docs/hack/functions/main.go
+++ b/docs/hack/functions/main.go
@@ -336,6 +336,15 @@ var Functions = []Function{
 		IsGlobal:    true,
 	},
 	{
+		Name:        "is_in",
+		Description: "Returns exit code 0 if the value of the first argument can be found in the second argument (second argument being a blank-separated list of strings e.g `\"bananas apples peaches\"`)",
+		Args:        `[value-1] [value-2]`,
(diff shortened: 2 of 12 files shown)


lizardruss (changes requested):
@jackwilsdon

Thanks for the PR!

It looks like there are some issues with our [function partials generation script](https://github.com/devspace-sh/devspace/blob/main/docs/hack/functions/main.go). The first being that it's not included in [gendocs.sh](https://github.com/devspace-sh/devspace/blob/main/gendocs.sh). The second is that it's not placing these in the correct category.

If you get a chance to look into those issues, that would be great! Otherwise I can take a look soon.

lizardruss:
@jackwilsdon Thank you so much! Yeah I think including it in this PR would be good. Adding `is_in` should be pretty straightforward. I'm just learning that the [text for that documentation is here](https://github.com/devspace-sh/devspace/blob/b5314931131d2feb6d4d231c2d7db85838289e8a/docs/hack/functions/main.go#L231-L433), and looks like it was updated but the partial generation was never run afterwards. Looks like `is_dependency` can be moved to the correct group by removing [this flag](https://github.com/devspace-sh/devspace/blob/b5314931131d2feb6d4d231c2d7db85838289e8a/docs/hack/functions/main.go#L318).

You've probably figured all this out, but I'm also just learning about this part of the document generation code 👍

jackwilsdon:
I've updated this PR to include the changes discussed :+1:

FabianKramm (approved):
@jackwilsdon thanks a lot for the PR! LGTM!

lizardruss (approved):

PR #2769, opened 2023-12-04
Draft for dependencies cache reuse within same deploy

### LINKED ISSUE #2768: Avoid uncessary dependency downloading for devspace deploy
Kinda. Devspace redownload dependencies within the same devspace deploy command run. Meaning with a complex enough dependency heirarchy the same repo could be git pulled 10+ times.

Draft PR: https://github.com/devspace-sh/devspace/pull/2769

Simplifying dependency hierarchy

Andrioden:
/kind enhancement

resolves #2768

This is an incomplete PR, my limited understanding of the go language discouraged me to continue as i could not see a pretty way forward to properly preserve the download history. I am also very uncertain how you would want it solved for your codebase.

So either take this an inspiration for what i would like, or maybe give some pointers on how I could move forward.


pkg/devspace/dependency/resolver.go
```diff
@@ -31,6 +31,7 @@ type ResolverInterface interface {
 // Resolver implements the resolver interface
 type resolver struct {
 	DependencyGraph *graph.Graph
```
lizardruss, line 33: I think you may be able to use the DependencyGraph to avoid re-downloading / re-pulling dependencies.

pkg/devspace/dependency/resolver.go
```diff
@@ -130,7 +131,7 @@ func (r *resolver) resolveRecursive(ctx devspacecontext.Context, basePath, paren
 			continue
 		}
 
-		dependencyConfigPath, err := util.DownloadDependency(ctx.Context(), basePath, dependencyConfig.Source, ctx.Log())
+		dependencyConfigPath, err := util.DownloadDependency(ctx.Context(), basePath, dependencyConfig.Source, ctx.Log(), r.DownloadedIds)
```
lizardruss, line 134: You might be able to skip this by doing a check similar to [this line](https://github.com/devspace-sh/devspace/pull/2769/files#diff-619d41cab5b9425e3f17b291d9018247b2ce6c2a7d9ec9d56e65a30602191bceR143).

lizardruss (changes requested):
There may be an existing data structure to use for tracking which dependencies have been resolved.

There's not currently an easy way to e2e test changes to dependency resolution. I'll bring it up for discussion and see if we want to add something to support it.

PR #2777, opened 2023-12-21
fix: use string hash instead of whole name

### LINKED ISSUE #2776: Devspace dev throws exception when the container name is too long and the `startContainer` is set to `true`
**What happened?** 

Devspace dev command is throwing an error when the container name is too long when `startContainer` is set to `true` because it's trying to inject the `restart-helper`:
```
[
  spec.template.annotations: Invalid value: "devspace.sh/restart-helper-super-very-long-name-super-very-long-name-super-very-long-name": name part must be no more than 63 characters,
  spec.template.spec.volumes[8].downwardAPI.fieldRef: Invalid value: "devspace.sh/restart-helper--super-very-long-name-super-very-long-name-super-very-long-name": name part must be no more than 63 characters
]
```

In this example, the container name is too long and the deployment creation is rejected by Kubernetes because the `fieldRef` value is too long ([Code](https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/services/podreplace/builder.go#L336)).

Ideally, Devspace trims the `fieldRef` if it's more than the characters limit or allow a customization of the value.

devspace.yaml:
```
version: v2beta1
name: nginx-k8s
pipelines:
  dev:
    run: |-
      start_dev app
dev:
  app:
    command:
      - /docker-entrypoint.sh
    args:
      - nginx
      - -g
      - daemon off;
    container: super-very-long-name-super-very-long-name-super-very-long-name
    labelSelector:
      app: nginx
    sync:
      - path: ./:/app
        disableDownload: true
        startContainer: true
    ports:
      - port: 8080:80
```
manifest.yaml:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1 # tells deployment to run 2 pods matching the template
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: super-very-long-name-super-very-long-name-super-very-long-name
          image: nginx:1.14.2
          ports:
            - containerPort: 80
```

**Local Environment:**  
- DevSpace Version: 6.3.7
- Operating System: mac
- ARCH of the OS: ARM64
**Kubernetes Cluster:**  
- Cloud Provider: aws 
- Kubernetes Version:
```
Client Version: v1.26.2
Kustomize Version: v4.5.7
Server Version: v1.27.6
```

FabianKramm:
/kind bugfix

resolves #2776


lizardruss (approved):
LGTM

PR #2789, opened 2024-01-12
restart helper: give container process some time to shutdown

### LINKED ISSUE #2788: Helper restart does not give container process the chance to shut down properly
**What happened?**  

When my container is restarted due to sync.onUpload.restartContainer, the container process is (almost) immediately killed. In my case, this prevents the watcher from cleaning up the subprocess running in another process group.

The restart helper should send SIGINT, SIGTERM and finally SIGKILL, with some time in-between the signals.

My devspace.yaml:
```
version: v2beta1
dev:
  mypod:
    # ...
    sync:
      - path: .:.
        onUpload:
          restartContainer: true
```

**Local Environment:**  
- DevSpace Version: 6.3.8
- Operating System: linux
- ARCH of the OS: AMD64
**Kubernetes Cluster:**  
- Kubernetes Version: v1.26.3+k3s1

maartendeprez:
/kind bugfix

resolves #2788

The helper restart command would wait for 2 microseconds (2000 nanoseconds, supposedly meant to be 2 seconds) between sending SIGINT and SIGKILL to the subcommand. Changed this to send SIGINT, SIGTERM and SIGKILL with five seconds in between, checking for process termination.

### DIFF
--- a/helper/util/restart_linux.go
+++ b/helper/util/restart_linux.go
@@ -11,8 +11,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/loft-sh/devspace/helper/util/stderrlog"
 	"github.com/loft-sh/devspace/pkg/devspace/build/builder/restart"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 type containerRestarter struct {
@@ -69,8 +71,23 @@ func (*containerRestarter) RestartContainer() error {
 	}
 
 	// kill the process group
-	_ = syscall.Kill(-pgid, syscall.SIGINT)
-	time.Sleep(2000)
-	_ = syscall.Kill(-pgid, syscall.SIGKILL)
+	procPath := "/proc/" + strconv.Itoa(pgid)
+
+	for _, sig := range []syscall.Signal{syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL} {
+		stderrlog.Infof("Sending %s signal...", sig.String())
+		err = syscall.Kill(-pgid, sig)
+		if err != nil {
+			return nil
+		}
+		err = wait.PollImmediate(time.Second, 5*time.Second, func() (done bool, err error) {
+			_, err = os.Stat(procPath)
+			return os.IsNotExist(err), nil
+		})
+		if err == nil {
+			return nil
+		}
+	}
+
+	stderrlog.Errorf("Timeout waiting for the process to terminate")
 	return nil
 }


helper/util/restart_linux.go
```diff
@@ -69,8 +70,45 @@ func (*containerRestarter) RestartContainer() error {
...
-	time.Sleep(2000)
+	
+	for i := 0; i < 5; i++ {
+		time.Sleep(time.Second)
+		_, err := os.Stat(path)
+		if os.IsNotExist(err) {
+			stderrlog.Infof("Process terminated!")
+			return nil;
+		}
+	}
```
lizardruss, line 85: Could  use `wait.PollImmediate` here instead of a loop. It's already [included in the helper it seems](https://github.com/devspace-sh/devspace/blob/main/helper/util/pingtimeout/ping_timeout.go#L33).

helper/util/restart_linux.go
```diff
@@ -69,8 +70,45 @@ func (*containerRestarter) RestartContainer() error {
...
+
+	stderrlog.Infof("Sending SIGTERM...")
```
lizardruss, line 87: You might consider moving the wait logic into a function. Something along the lines of:
```golang
stderrlog.Infof("Sending SIGINT...")
_ = syscall.Kill(-pgid, syscall.SIGINT)
if done, err := waitForTerminated(pgid); err != nil {
  return err
} else if done {
  return nil
}
```

lizardruss (commented):
Added some ideas for reducing the repetition.

helper/util/restart_linux.go
maartendeprez: Updated. According to the [docs](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate), `wait.PollImmediate` is deprecated, but the version already included in the helper does not have the replacement function yet, so i used it anyway. Would it be better to update the import?

Alternatively, we could wait for the process using the [pidfd](https://github.com/oraoto/go-pidfd) interface. Supposedly this would be slightly more efficient, but also, more importantly, it would avoid a race condition in case the pid would be reused between sending the signal and polling for termination. It would require an extra import and might make the helper somewhat less portable though (the syscall appeared in Linux 5.3).

```go
func KillPidFd(pid int) {

        fmt.Printf("Opening process %d\n", pid)

        fd, err := pidfd.Open(pid, 0)
        if err != nil {
                fmt.Printf("Error: %s\n", err.Error())
                return
        }

        for _, sig := range []syscall.Signal{syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL} {
                fmt.Printf("Sending %s signal...\n", sig.String())

                err = fd.SendSignal(sig, 0)
                if err != nil {
                        fmt.Printf("Error: %s\n", err.Error())
                        return
                }

                fds := []unix.PollFd{{Fd: int32(fd), Events: unix.POLLIN}}
                n, _ := unix.Poll(fds, 5000 /* milliseconds */)

                if n > 0 {
                        fmt.Printf("Process terminated!\n")
                        return
                }
        }

        fmt.Printf("Timeout!\n")

}
```

lizardruss: @maartendeprez 

This is already a good improvement over the original code. I think we can wait to upgrade to the replacement for `wait.PollImmediate`, so that it can upgraded outside the helper as well.

Regarding the `pidfd` interface, if the race condition has been observed in your testing, and happens frequently enough, then we can consider adding it. If it hasn't been observed then we can skip for now.

lizardruss (approved):
LGTM
Resolves ENG-2674

PR #2798, opened 2024-01-30
fix: treat pipeline flag defaults consistently when used as command f…

lizardruss:
…lags or pipeline flags

/kind bugfix

resolves ENG-2735

Fixed an issue where DevSpace would treat pipeline flag default values differently when used as command flags vs pipeline flags.


89luca89 (approved):
LGTM!

PR #2815, opened 2024-03-12
feat(ci): add License checker bot

89luca89:
/kind bugfix
/kind test

DevSpace CNCF Compliance: Add GH action license checker

Resolve Eng-2966


docs/sidebars.js
```diff
@@ -206,5 +206,9 @@ module.exports = {
       ],
     },
     "plugins/README",
+    {
+        type: "doc",
+        id: "licenses/devspace",
+    },
```
lizardruss, line 212: Interesting that this was added to the docs. Is that one of the CNCF requirements?

lizardruss (commented):
LGTM, but had some questions.

I didn't expect to see a page added to the hosted docs, so just curious what the context around that is.

Also should we update [the license part](https://github.com/devspace-sh/devspace?tab=readme-ov-file#license) of the repository README?

docs/sidebars.js
89luca89: Nop but having this available I thought it was a nice addition (also Vcluster has this)

89luca89:
> Also should we update [the license part](https://github.com/devspace-sh/devspace?tab=readme-ov-file#license) of the repository README?

I think it should be fine like this, that part is about Devspace's license, the other are for dependencies, I prefer to keep it clean like this, wdyt?

docs/sidebars.js
lizardruss: Gotcha. Yeah I'm curious if that page will stay since we're redoing the vcluster docs. No problem keeping it, was mostly just curious.

PR #2864, opened 2024-06-10
fix remote buildkit error: [415: Unsupported Media Type]

wurenny:
# Issue describe
since legacy kube api not support server-side apply, when use remote buildkit will get the error:
```shell
local-registry: Starting Local Image Registry
build_images: build images: create builder: get or create local registry: ensure deployment: 415: Unsupported Media Type
fatal exit status 1
```
this pr will resolve the compatibility issue

# The pr summary
/kind bugfix

fix remote buildkit error: [415: Unsupported Media Type], compatibility legacy kube api server which not support server-side apply

# repreduce
```shell
$ curl -Lo ~/bin/devspace-linux-amd64-6.3.12 https://github.com/devspace-sh/devspace/releases/download/v6.3.12/devspace-linux-amd64
$ chmod u+x ~/bin/devspace-linux-amd64-6.3.12

$ ~/bin/devspace-linux-amd64-6.3.12  version
DevSpace version : 6.3.12

$ kubectl version --short
Client Version: v1.26.7
Kustomize Version: v4.5.7
Server Version: v1.15.11

$ devspace-linux-amd64-6.3.12 build --skip-push
info Using namespace 'devspace'
info Using kube context 'devspace@dev'
Couldn't find a running docker daemon. Will fallback to local registry
local-registry: Starting Local Image Registry
build_images: build images: create builder: get or create local registry: ensure deployment: 415: Unsupported Media Type
fatal exit status 1
```


pkg/devspace/build/localregistry/deployment.go
```diff
@@ -58,11 +58,16 @@ func (r *LocalRegistry) ensureDeployment(ctx devspacecontext.Context) (*appsv1.D
 		return nil, err
 	}
 
+	if !ctx.KubeClient().SupportServerSideApply() {
```
lizardruss, line 61: IIRC, server side apply can also be disabled. Perhaps handling the 415 response and falling back to a Patch operation would be an option that would adapt to more version / configuration combinations?

lizardruss (changes requested):
Thanks for the PR!

I might prefer a non-version based approach to solve this. See my comments. If it's not practical, we can discuss.

pkg/devspace/build/localregistry/deployment.go
wurenny: Yes, I proved it, and I would fix it with non-version based approach.
```shell

$ kubectl version
Client Version: v1.30.0
Server Version: v1.18.19

$ kubectl -n kube-system get po -l component=kube-apiserver -o yaml | grep -i serversideapply
      - --feature-gates=ServerSideApply=false

$ kubectl apply --server-side=true -f - <<EOF
> apiVersion: v1
> kind: ConfigMap
> metadata:
>   name: ssa-cm
>   namespace: default
> data:
>   foo: bar
> EOF
error: Server-side apply not available on the server: (415: Unsupported Media Type)
```

lizardruss (approved):
LGTM

PR #2866, opened 2024-06-12
fix: do not wait for initContainer to exit when it is a sidecar

### LINKED ISSUE #2865: devspace open will wait for a init container which is a sidecar
**What happened?**  
In kubernetes 1.29 the Sidecar feature became general available. This means you have a init container which has a `restartPolicy: Always` runs forever, effectively as a sidecar. This is what istio uses to prevent a lot of [issues](https://istio.io/latest/blog/2023/native-sidecars/#sidecar-woes)

The problem is devspace open will wait for the init container to exit, therefore, devspace dev will never continue.

If `restartPolicy` is defined in the init container it should just continue.

Use Kubernetes 1.29 or higher. And use a manifest like https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

devspace open wil not continue

nberlee:
fixes #2865

/kind bugfix

resolves #2865 

waits for native sidecar containers (init container)


lizardruss (changes requested):
@nberlee 

It looks like the updated golang version will require changes to the GitHub Actions. Is this something you feel comfortable looking into?

lizardruss (approved):

PR #2902, opened 2024-10-04
fix: devspace ui not working due to incorrect dependencies

### LINKED ISSUE #2899: Devspace ui fails for v6.3.13
**What happened?**  
I ran "devspace ui" from cli.
In cli everything as expected:

```
info Using namespace 'test-namespace'
info Using kube context 'minikube'
info Start listening on http://localhost:8090
```
When I open  http://localhost:8090 in the browser (chrome) nothing is presented. In the Chrome Dev Tools console this error is shown:

```
 index.tsx:6 Uncaught TypeError: o.render is not a function
    at Module.<anonymous> (index.tsx:6:1)
    at i (bootstrap:19:22)
    at Object.<anonymous> (main.a9eb55d6.js:1:298982)
    at i (bootstrap:19:22)
    at bootstrap:83:10
    at main.a9eb55d6.js:1:913
```

Devspace ui opens in the browser.

**How can we reproduce the bug?** 
install devspace version 6.3.13
run in cli: devspace ui and open link with browser

**Local Environment:**  
- DevSpace Version: 6.3.13
- Operating System: windows | WSL 

**Current Fix**
use Devspace version: 6.3.12

lizardruss:
/kind bugfix

resolves #2899
resolves DSP-92

Fixed an issue where DevSpace UI would not work due to an error

Some other dependencies were downgraded due to engine incompatibility with node 12, which is current used to bundle the react app.


zerbitx (approved):
lgtm

PR #2924, opened 2025-01-16
fix: Removes negative lookahead from regex

zerbitx:
/kind bugfix

resolves #DSP-101

### DIFF
--- a/pkg/devspace/configure/image.go
+++ b/pkg/devspace/configure/image.go
@@ -10,11 +10,11 @@ import (
 	"path"
 	"regexp"
 	"strings"
-
+	
 	"github.com/loft-sh/devspace/pkg/util/encoding"
 	"github.com/loft-sh/utils/pkg/command"
 	"github.com/sirupsen/logrus"
-
+	
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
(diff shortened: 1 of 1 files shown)


pkg/devspace/configure/image.go
```diff
@@ -174,7 +174,7 @@ func (m *manager) AddImage(imageName, image, projectNamespace, dockerfile string
 				registryHostname, err = m.log.Question(&survey.QuestionOptions{
 					Question:               "Please provide the registry hostname without the image path (e.g. gcr.io, ghcr.io, ecr.io)",
 					DefaultValue:           "gcr.io",
-					ValidationRegexPattern: "^(?!-)[a-z0-9-]{1,63}(\\.[a-z0-9-]{1,63})*$",
+					ValidationRegexPattern: "^[a-z0-9][a-z0-9-]{0,62}(\\.[a-z0-9-]{1,63})*$",
```
lizardruss, line 177: only nit: use backtick string to avoid the double escape `\\.`:
```suggestion
					ValidationRegexPattern: `^[a-z0-9][a-z0-9-]{0,62}(\.[a-z0-9-]{1,63})*$`,
```

lizardruss (approved):
lgtm

PR #2937, opened 2025-02-14
Allow user to show/hide internal commands in ui

### LINKED ISSUE #2923: The ui shows internal commands when it maybe shouldn't
**What happened?**  
When I visit the commands page of the ui i can see and execute internal commands

I would expect internal commands to not appear at all or at least to not be executable

Define an internal command and check the commands page
My devspace.yaml:
```
version: v2beta1
...
```

**Local Environment:**  
- DevSpace Version: 6.3.14
- Operating System: windows | linux | mac
- ARCH of the OS: AMD64 | ARM64 | i386
**Kubernetes Cluster:**  
- Cloud Provider: google | aws | azure | other
- Kubernetes Version: 
  Client Version: v1.31.0
  Kustomize Version: v5.4.2
  Server Version: v1.30.6+k3s1

demetris-manikas:
**What issue type does this pull request address?
/kind feature

resolves #2923

Allow users to toggle the visibility of internal commands in the ui


ui/src/pages/commands/commands.tsx
```diff
@@ -81,11 +84,18 @@ class Commands extends React.PureComponent<Props, State> {
   };
 
   render() {
+    let commands = this.props.devSpaceConfig.config.commands;
```
zerbitx, line 87: I checked out your branch, fired it up and got the folllowing errors in the console, preventing the page to load

ds-console-error

demetris-manikas: Sorry about that.
Trusting typescript is always dodgy.

Obviously the definitions in 
https://github.com/devspace-sh/devspace/blob/a633a2b35d4564b45c5fecfbb4646ccb331e4ae7/ui/src/contexts/withDevSpaceConfig/withDevSpaceConfig.tsx

are not proper.
Give it another try.

demetris-manikas:
Hi.
Thanks for giving devspace to the world.
It is a great product.
It seems that you either have no time or no interest in this.
I thought you asked for a PR because you wanted people to get involved.
Ignoring for so long does not help.
The changes are minimal.
If you wish you can take it from here.
I won't bother with matter any longer.
Thanks

PR #2973, opened 2025-07-03
fix: prevent using "<nil>" as default value

### LINKED ISSUE #2971: Default for question vars is "<nil>"
**What happened?**  
The default value for question vars is "<nil>". This causes issues when trying to make a variable required using `validationPattern` (see attached `devspace.yaml`).

The default value to be an empty string.

1. `devspace deploy` the attached `devspace.yaml`

My devspace.yaml:
```
version: v2beta1
vars:
  EXAMPLE:
    question: Example
    validationPattern: .+
    validationMessage: An access token is required
```

The workaround is to also set `default: ""`. Interestingly enough, password variables (`password: true`) do default to an empty string from what I can tell.

**Local Environment:**  
- DevSpace Version: [use `devspace --version`] 6.3.15
- Operating System: macOS
- ARCH of the OS: ARM64

lizardruss:
/kind bugfix

resolves #2971

Fixed an issue where DevSpace would default a string var to "`<nil>`"

### DIFF
--- a/pkg/devspace/config/loader/variable/undefined_variable.go
+++ b/pkg/devspace/config/loader/variable/undefined_variable.go
@@ -99,7 +99,7 @@ func getParams(variable *latest.Variable) *survey.QuestionOptions {
 			params.IsPassword = true
 		}
 
-		if variable.Default != "" {
+		if variable.Default != nil && variable.Default != "" {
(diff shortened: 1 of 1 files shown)


zerbitx (approved):
lgtm

PR #2983, opened 2025-08-04
Dependabot pr combined

zerbitx:
/kind enhancement

resolves #DSP-117

I went through the dependabot PRs and incorporated all those that worked without major changes and closed those that had bigger issues like API changes or breaking a bunch of peer dependencies after major version package updates. 


pkg/devspace/build/builder/helper/helper_test.go
```diff
@@ -4,33 +4,47 @@ import (
...
-func (d *fakeDockerClient) DockerAPIClient() dockerclient.CommonAPIClient {
+func (d *fakeDockerClient) DockerAPIClient() dockerclient.APIClient {
 	return &fakeDockerAPIClient{}
 }
 
 type fakeDockerAPIClient struct {
 	dockerclient.CommonAPIClient
 }
 
+// TODO: (Ryan) see if these are needed for the tests
```
lizardruss, line 35: were they?

lizardruss (approved):
lgtm, as long as the UI was tested. We don't have an automated test for it, and we've had cases where a release went out with a broken UI. LMK if you need some pointers on testing it.

PR #3103, opened 2026-01-15
feat(imports): add mergeStrategy field to support deep merge

JordanGoasdoue:

/kind feature

Adds an opt-in `mergeStrategy` field on imports to support recursive deep merge of config sections.

**Problem:**
When importing a config (e.g., from a git catalog), the current implementation only does a shallow merge at the first level. If both the imported config and the local config define the same section (like `dev.api`), nested maps are completely replaced instead of being merged.

Example:
```yaml
# Catalog
dev:
  api:
    labelSelector:
      app: catalog-app
      version: v1
    command: ["/bin/bash"]

# Local config
dev:
  api:
    labelSelector:
      app: my-app
```

**Before:** Local `dev.api` replaces catalog entirely, loses `version: v1`, `command`, etc.
**After (with `mergeStrategy: deepMerge`):** Maps are deep merged, keeps both `app: my-app` and `version: v1`

**Solution:**
A new `mergeStrategy` field on each import entry, with two values:

- `shallowMerge` (default, current behavior): only top-level keys within each section are merged. Existing keys in the local config take precedence entirely.
- `deepMerge` (opt-in): maps are merged recursively. Arrays and scalars in the local config still take precedence.

```yaml
imports:
  - git: https://github.com/org/catalog.git
    subPath: devspace.yaml
    mergeStrategy: deepMerge
```

**Merge rules when `deepMerge` is enabled:**
| Type | Behavior |
|---|---|
| Maps | Recursively merged (local keys override on conflict, catalog keys preserved otherwise) |
| Arrays | Local replaces catalog entirely |
| Scalars | Local overrides catalog |

This enables the common use case where organizations maintain a shared catalog in git, and projects only need to override specific values (like `labelSelector`) while inheriting everything else (ports, sync config, commands, etc.).

Imports now support a `mergeStrategy` field. Set `mergeStrategy: deepMerge` on an import to enable recursive map merging, allowing projects to inherit shared catalog configs while only overriding specific nested values. Default behavior (`shallowMerge`) is unchanged.

- **Fully backward compatible**: default is `shallowMerge`, which preserves the existing behavior exactly. No existing config is affected.
- **Mutually exclusive fields are safe**: since deep merge is opt-in, users who don't use it are not affected. When `deepMerge` combines fields that shouldn't coexist, the existing validations in `validate.go` catch it properly. This is tested explicitly:
  - `deployments`: `helm` + `kubectl` on the same deployment triggers `"deployments[x].kubectl and deployments[x].helm cannot be used together"`
  - `dev`: `imageSelector` + `labelSelector` on the same dev pod triggers `"dev.x: image selector and label selector cannot be used together"`
  - Both cases are also tested with `shallowMerge` to prove no regression (the conflicting field from the catalog is simply ignored)
- Comprehensive test coverage:
  - 11 integration tests via `TestImportsDeepMerge` (deep merge scenarios, default shallow, explicit shallow, mutually exclusive field validation in both `deployments` and `dev` sections)
  - 7 unit tests for `deepMerge` (edge cases: empty maps, 3-level nesting, type mismatches, nil values)
  - 4 unit tests for `shallowMerge`
- Schema files (`devspace-schema.json`, `config-openapi.json`) regenerated, IDE autocomplete works out of the box
- All existing tests pass

### DIFF
--- a/devspace-schema.json
+++ b/devspace-schema.json
@@ -2021,6 +2021,14 @@
           ],
           "description": "Enabled specifies if the given import should be enabled"
         },
+        "mergeStrategy": {
+          "type": "string",
+          "enum": [
+            "shallowMerge",
+            "deepMerge"
+          ],
+          "description": "MergeStrategy specifies how the imported config should be merged (e.g. shallowMerge or deepMerge)"
+        },
         "path": {
           "type": "string",
           "description": "Path is the local path where DevSpace can find the artifact.\nThis option is mutually exclusive with the git option.",
--- a/docs/schemas/config-openapi.json
+++ b/docs/schemas/config-openapi.json
@@ -1045,6 +1045,14 @@
                 "type": "boolean",
                 "description": "Enabled specifies if the given import should be enabled"
               },
+              "mergeStrategy": {
+                "type": "string",
+                "enum": [
+                  "shallowMerge",
+                  "deepMerge"
+                ],
+                "description": "MergeStrategy specifies how the imported config should be merged (e.g. shallowMerge or deepMerge)"
+              },
               "path": {
                 "type": "string",
                 "description": "Path is the local path where DevSpace can find the artifact.\nThis option is mutually exclusive with the git option.",
--- a/pkg/devspace/config/loader/imports.go
+++ b/pkg/devspace/config/loader/imports.go
@@ -118,13 +118,7 @@ func ResolveImports(ctx context.Context, resolver variable.Resolver, basePath st
 			if mergedMap[section] == nil {
 				mergedMap[section] = map[string]interface{}{}
 			}
-
-			for key, value := range sectionMap {
-				_, ok := mergedMap[section].(map[string]interface{})[key]
-				if !ok {
-					mergedMap[section].(map[string]interface{})[key] = value
-				}
-			}
+			deepMerge(mergedMap[section].(map[string]interface{}), sectionMap)
 		}
 
 		// resolve the import imports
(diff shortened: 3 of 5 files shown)


zerbitx:
Looking at this again, it looks like it will introduce some regressions with fields that shouldn't exist together like [helm](https://www.devspace.sh/docs/configuration/deployments/helm/https://www.devspace.sh/docs/configuration/deployments/helm/) and [kubectl](https://www.devspace.sh/docs/configuration/deployments/kubectl/)

JordanGoasdoue:
Good point @zerbitx, you were right. I've reworked it. Deep merge is now opt-in via a `mergeStrategy` field on each import. Default stays `shallowMerge`, so no behavior change for existing configs.

```yaml
imports:
  - path: catalog.yaml
    mergeStrategy: deepMerge
```

I added tests for the mutually exclusive fields you flagged. When `deepMerge` combines fields that shouldn't coexist, the existing validations catch it:
- `helm` + `kubectl` in deployments
- `imageSelector` + `labelSelector` in dev

Both cases also tested with `shallowMerge` to confirm no regression.

Let me know what you think!

zerbitx:
Thanks for the updates.  This is looking good.  I only have a short list of asks:

 - Run `go run ./docs/hack/config/partials` to generate the docs for the new field
 - Add a check for validity of the new field in [this loop](https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/config/loader/imports.go#L69)
 - Add a test for 👆 by having a config in a test case with something like `mergeStrategy: invalid`

PR #3105, opened 2026-01-15
fix: restore logging for dependency operations

JordanGoasdoue:

/kind bugfix

Fixes dependency operations running with completely discarded logs (including errors and fatal messages), making failures impossible to troubleshoot.

**Problem:**

Dependency resolution (git clone, pull, etc.) was using a discard logger, causing all log messages to be lost, including errors and fatal failures. This happened because:

1. The logger was initialized before command-line flags were parsed
2. During config loading (which includes dependency resolution), a discard logger was used
3. Even `--debug` and `--log-level` flags had no effect

**Impact:**
- Git clone failures: silent (no error shown)
- Git pull failures: silent (no error shown)
- Network issues: silent (no error shown)
- Any dependency error: completely invisible
- Impossible to troubleshoot dependency issues

**Solution:**

Parse command-line flags early in the boot sequence (before config loading) to configure the logger with the correct level, then use the factory logger (which properly outputs) instead of discard logger during dependency operations.

**Changes:**

- Modified `BuildRoot()` to parse flags before calling `parseConfig()`
- Added `configureLogLevel()` helper function to centralize log level configuration
- Modified `parseConfig()` to use factory logger instead of `log.Discard`
- Ensured `SetDevSpaceRoot()` and `LoadWithParser()` use the properly configured logger

Fixed dependency operations running with discarded logs, restoring visibility into git clone/pull operations and making troubleshooting possible.

**Note:** This PR works in conjunction with #3104  (auto-reclone feature). The "After" example below shows both fixes working together:
- This PR enables the logs to be displayed
- PR #3104 provides the auto-reclone recovery mechanism

### Testing

Tested with a git dependency that has rewritten history (force-pushed branch).

**Before (logs discarded):**
```bash
devspace build
info Using namespace 'xxx'
info Using kube context 'xxx'
Building...
# No information about dependency pull failure
# Silently continues with stale cached version
```

**After (logs properly displayed):**
```bash
devspace build
info Git pull failed for dependency 'xxx.'
[...]
info Attempting automatic recovery by re-cloning
done Successfully recovered dependency 'https://github.com/xxx.git'
info Using namespace 'xxx'
info Using kube context 'xxx'
Building...

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -10,13 +10,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/loft-sh/devspace/pkg/devspace/kill"
 	"github.com/mgutz/ansi"
 
 	"github.com/loft-sh/devspace/pkg/devspace/config/loader"
-	"github.com/loft-sh/devspace/pkg/devspace/config/loader/variable/expression"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
 	"github.com/loft-sh/devspace/pkg/devspace/env"
+	"github.com/loft-sh/devspace/pkg/devspace/kill"
 	"github.com/loft-sh/devspace/pkg/util/log"
 	"github.com/loft-sh/devspace/pkg/util/message"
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/root.go [hunk 1]
```diff
@@ -191,6 +190,20 @@ func BuildRoot(f factory.Factory, excludePlugins bool) *cobra.Command {
...
 	}
 
+	rootCmd := NewRootCmd(f)
+	rootCmd.SetHelpCommand(&cobra.Command{
+		Use:    "no-help",
+		Hidden: true,
+	})
+
+	persistentFlags := rootCmd.PersistentFlags()
+	globalFlags = flags.SetGlobalFlags(persistentFlags)
```
zerbitx, line 200: Parsing this early will affect commands that set `DisableFlagParsing` to `true` like the run command, so if say you wanted to do `devspace run my-script --namespace demo`, where you intend the namespace flag only for the script, but not for devspace, this changes that behavior

PR #3205, opened 2026-04-05
fix: noWatch sync no longer prevents SSH injection

### LINKED ISSUE #3005: Adding `noWatch: true` prevents ssh injection
**What happened?**  
I configured a devspace.yaml, but for some reason the SSH server injection never happened, despite the `ssh:` section being present and all similar devspace.yamls working fine.

SSH server injection happens and SSH server and forwarding starts, i.e. the `devspace dev` command yields the following lines at the end:

```
dev:my-project ssh   Port forwarding started on: 11791 -> 8022
dev:my-project ssh   Use 'ssh my-project.devspace' to connect via SSH
```

Through painful hours of trial and error I figured out that the bug was caused by adding a sync configuration that had `noWatch: true` set.

My devspace.yaml:
```yaml
---
version: v2beta1
name: my-project

vars:
  DEVSPACE_ENV_FILE: ".env"

# deploy a simple pod
deployments:
  umsatzprognose-minimal-devspace:
    kubectl:
      manifests:
        - k8s-deployment.yaml

dev:
  my-project:
    labelSelector:
      app: devspace
      project: my-project
      environment: my-env
    container: devspace
    workingDir: /home/jovyan/
    ssh:
      localHostname: my-project.devspace
    ports:
      - port: 9000:9000
    sync:
      - path: ./pyproject.toml:/home/jovyan/pyproject.toml
        file: true
        #noWatch: true  # random devspace bug: enabling this prevents ssh from starting. ???
```

**Local Environment:**  
- DevSpace Version: 6.3.15
- Operating System: windows
- ARCH of the OS: AMD64
**Kubernetes Cluster:**  
- Cloud Provider: aws
- Kubernetes Version:
  Client Version: v1.34.1
  Kustomize Version: v5.7.1
  Server Version: v1.29.14+a6b193c

raman1236:
## Problem

When a sync config has `noWatch: true`, SSH server injection never starts. The SSH section in `devspace.yaml` is completely ignored when any sync entry has `noWatch: true`.

## Root Cause

The dev mode starts services in sequence: **sync → port forwarding → SSH**. SSH only starts after sync initialization completes (`<-syncDone`).

When `noWatch: true`:

1. `StartSync` creates a cancelable context for that sync entry
2. The sync goroutine starts, completes initial sync, and returns
3. `defer cancel()` fires, cancelling the context
4. Meanwhile, `startWithWait()` had registered a `RestartOnError` goroutine via `parent.Go()` that listens on `ctx.Context().Done()`
5. The context cancellation triggers this handler, which calls `syncStop()` → `parent.Kill(nil)`
6. The parent tomb enters a dying state
7. SSH (which starts after `<-syncDone`) either cannot start or is immediately terminated because the tomb is dying

## Fix

Set `RestartOnError` to `false` when `NoWatch` is `true`. A one-shot sync has no need for restart-on-error handling since it completes after initial sync. This prevents the restart handler goroutine from being registered, which in turn prevents it from killing the parent tomb when the cancelled context fires.

## Change

In `pkg/devspace/services/sync/sync.go`, `startSync()`:
```go
// Before:
RestartOnError: true,

// After:
RestartOnError: !syncConfig.NoWatch,
```

Fixes #3005

### DIFF
--- a/pkg/devspace/services/sync/controller.go
+++ b/pkg/devspace/services/sync/controller.go
@@ -239,7 +239,15 @@ func (c *controller) startWithWait(ctx devspacecontext.Context, options *Options
 				ctx.Log().Errorf("Sync error: %v", err)
 				syncStop(ctx, client, options, parent)
 			case <-onDone:
-				syncDone(ctx, options, parent)
+				// noWatch: initial sync completed normally.
+				// Stop the client and emit the stop:sync lifecycle event,
+				// but do NOT kill the parent tomb — SSH, port-forwarding and
+				// other services registered in the same parent must continue.
+				client.Stop(nil)
+				hook.LogExecuteHooks(ctx.WithLogger(options.SyncLog), map[string]interface{}{
+					"sync_config": options.SyncConfig,
+				}, hook.EventsForSingle("stop:sync", options.Name).With("sync.stop")...)
+				ctx.Log().Debugf("Stopped sync %s", options.SyncConfig.Path)
 			}
 			return nil
 		})
--- a/pkg/devspace/services/sync/controller.go
+++ b/pkg/devspace/services/sync/controller.go
@@ -226,6 +226,23 @@ func (c *controller) startWithWait(ctx devspacecontext.Context, options *Options
 			}
 			return nil
 		})
+	} else {
+		parent.Go(func() error {
+			select {
+			case <-ctx.Context().Done():
+				syncStop(ctx, client, options, parent)
+			case err := <-onError:
+				hook.LogExecuteHooks(ctx.WithLogger(options.SyncLog), map[string]interface{}{
+					"sync_config": options.SyncConfig,
+					"ERROR":       err,
+				}, hook.EventsForSingle("error:sync", options.Name).With("sync.error")...)
+				ctx.Log().Errorf("Sync error: %v", err)
+				syncStop(ctx, client, options, parent)
+			case <-onDone:
+				syncDone(ctx, options, parent)
+			}
+			return nil
+		})
 	}
 
 	return nil
--- a/pkg/devspace/services/sync/controller_test.go
+++ b/pkg/devspace/services/sync/controller_test.go
@@ -1,8 +1,12 @@
 package sync
 
 import (
-	"gotest.tools/assert"
 	"testing"
+	"time"
+
+	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
+	"github.com/loft-sh/devspace/pkg/util/tomb"
+	"gotest.tools/assert"
 )
 
 type parseSyncPathTestCase struct {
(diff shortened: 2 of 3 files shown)


zerbitx:
It looks like this will disable [this code path](https://github.com/devspace-sh/devspace/blob/main/pkg/devspace/services/sync/controller.go#L182) preventing sync:stop from being emitted.

raman1236:
@zerbitx Good catch! You're right that disabling `RestartOnError` also disables the `<-onDone` handler which emits `sync:stop`.

However, I believe for `noWatch` mode, the current code path is actually *the source of the bug* — here's why:

1. For noWatch, `StartSync` creates a cancelable context and defers `cancel()`
2. When `startSync()` returns after initial sync completes, `defer cancel()` fires
3. This cancels the context, triggering `<-ctx.Context().Done()` in the RestartOnError goroutine
4. That calls `syncStop()` → `parent.Kill(nil)` → kills the parent tomb
5. SSH (which starts after sync) can't start because the tomb is dying

The `sync:stop` emission in the old code is actually a *side effect* of this bug — it fires because the context cancellation triggers `syncStop()`, not because the sync truly "stopped."

That said, you're right that we should still emit `sync:stop` for noWatch completions. I'll push an update that adds a lighter goroutine for noWatch mode that only emits the `sync:stop` hook without calling `parent.Kill(nil)`:

```go
if options.RestartOnError {
    // existing handler (for watch mode)...
} else {
    // For noWatch: emit sync:stop when context is cancelled (sync complete)
    // but do NOT kill the parent tomb
    parent.Go(func() error {
        select {
        case <-ctx.Context().Done():
        case <-onDone:
        }
        client.Stop(nil)
        hook.LogExecuteHooks(ctx.WithLogger(options.SyncLog), map[string]interface{}{
            "sync_config": options.SyncConfig,
        }, hook.EventsForSingle("stop:sync", options.Name).With("sync.stop")...)
        return nil
    })
}
```

This preserves the `sync:stop` event while avoiding the parent tomb kill that breaks SSH. Let me push this fix.

raman1236:
Friendly ping — I've addressed @zerbitx's review feedback: the fix now preserves the `sync:stop` lifecycle event while still preventing SSH injection from being killed. The else branch handles noWatch syncs with a lighter goroutine. Would appreciate another review. Thanks!

pkg/devspace/services/sync/controller.go
```diff
@@ -226,6 +226,21 @@ func (c *controller) startWithWait(ctx devspacecontext.Context, options *Options
 			}
 			return nil
 		})
+	} else {
+		// For noWatch syncs: emit sync:stop when context is cancelled or sync
+		// completes, but do NOT kill the parent tomb (to avoid breaking SSH).
+		parent.Go(func() error {
+			select {
+			case <-ctx.Context().Done():
+			case <-onDone:
```
zerbitx, line 241: Can we add a case for [onError](https://github.com/devspace-sh/devspace/pull/3205/changes?w=1#diff-8dad5ae96232251c6b27a9fdc804028881f5f826566f453c994f5e31ccf5972fL92) as well with `hook.LogExecuteHooks(...`

pkg/devspace/services/sync/controller.go
zerbitx, line 159: I think we might also want to add a `if !options.SyncConfig.NoWatch` guard here.  I'll have to think on that more, just noting it here so I don't forget.

zerbitx:
@raman1236 Thanks again for this PR, and the follow up changes and explanation.  Its looking good.  I've added a couple comments and if you could add some testing we should be good to merge.

PR #3217, opened 2026-04-16
fix: recover from panic in IsMinikubeKubernetes when kubeconfig extensions are plain strings

### LINKED ISSUE #3218: Panic in IsMinikubeKubernetes when kubeconfig extensions contain plain string values
**What happened?**

`devspace build` panics with `reflect.Set: value of type string is not assignable to type map[string]interface {}` when the kubeconfig contains extensions with plain string values (e.g. from Teleport).

```
panic: reflect.Set: value of type string is not assignable to type map[string]interface {}

goroutine 222 [running]:
reflect.Value.assignTo(...)
k8s.io/apimachinery/pkg/runtime.toUnstructured(...)
	k8s.io/apimachinery@v0.29.0/pkg/runtime/converter.go:660
k8s.io/apimachinery/pkg/runtime.(*unstructuredConverter).ToUnstructured(...)
	k8s.io/apimachinery@v0.29.0/pkg/runtime/converter.go:586
github.com/loft-sh/devspace/pkg/devspace/kubectl.IsMinikubeKubernetes(...)
	pkg/devspace/kubectl/util.go:258
github.com/loft-sh/devspace/pkg/devspace/kubectl.IsLocalKubernetes(...)
	pkg/devspace/kubectl/util.go:217
github.com/loft-sh/devspace/pkg/devspace/build/builder/buildkit.(*Builder).BuildImage(...)
	pkg/devspace/build/builder/buildkit/buildkit.go:111
```

The kubeconfig is valid — the v1 spec defines `Extension` as `runtime.RawExtension`, which accepts any JSON value including plain strings. Teleport generates extensions like:

```yaml
clusters:
- cluster:
    extensions:
    - extension: mycompany.teleport.sh
      name: kubeconfig.teleport.dev/profile-name
```

`IsMinikubeKubernetes` calls `runtime.DefaultUnstructuredConverter.ToUnstructured()` on these extensions, which panics because it expects a struct/map, not a string.

`devspace build` should not panic on valid kubeconfig files. The `ToUnstructured` call should either be guarded with `recover()`, or replaced with direct type assertion on `*runtime.Unknown` and `json.Unmarshal` of its raw bytes.

**How can we reproduce the bug?**

Use any kubeconfig with plain-string cluster extensions (Teleport generates these by default) and run `devspace build`.

**Local Environment:**
- DevSpace Version: 6.3.20
- Operating System: macOS
- ARCH: ARM64

**Kubernetes Cluster:**
- Cloud Provider: AWS (via Teleport proxy)
- Kubernetes Version: v1.31

Workaround: strip extensions from the kubeconfig before invoking devspace:

```bash
kubectl config view --raw --minify | python3 -c "
import sys, yaml
config = yaml.safe_load(sys.stdin)
for c in config.get('clusters', []): c.get('cluster', {}).pop('extensions', None)
for c in config.get('contexts', []): c.get('context', {}).pop('extensions', None)
yaml.dump(config, sys.stdout, default_flow_style=False)
" > /tmp/stripped-kubeconfig.yaml
KUBECONFIG=/tmp/stripped-kubeconfig.yaml devspace build
```

benmosher:
## Problem

When using a Teleport-managed kubeconfig, `devspace build` (and any pipeline that calls `build_images`) panics and crashes:

```
panic: reflect.Set: value of type string is not assignable to type map[string]interface {}

goroutine 123 [running]:
k8s.io/apimachinery/pkg/runtime.toUnstructured(...)
    vendor/k8s.io/apimachinery/pkg/runtime/converter.go:660
k8s.io/apimachinery/pkg/runtime.(*unstructuredConverter).ToUnstructured(...)
    vendor/k8s.io/apimachinery/pkg/runtime/converter.go:586
github.com/loft-sh/devspace/pkg/devspace/kubectl.IsMinikubeKubernetes(...)
    pkg/devspace/kubectl/util.go:258
github.com/loft-sh/devspace/pkg/devspace/kubectl.IsLocalKubernetes(...)
    pkg/devspace/kubectl/util.go:217
github.com/loft-sh/devspace/pkg/devspace/build/builder/buildkit.(*Builder).BuildImage(...)
    pkg/devspace/build/builder/buildkit/buildkit.go:111
```

## Root cause

Teleport writes kubeconfig cluster extensions as plain YAML strings, e.g.:

```yaml
clusters:
  - cluster:
      extensions:
        - name: teleport.kube.name
          extension: some-cluster-name   # <-- bare string, not an object
```

The Kubernetes spec says extension values should be structured objects. `runtime.DefaultUnstructuredConverter.ToUnstructured()` uses reflection and assumes the extension value is a struct/map — it panics (rather than returning an error) when it encounters a plain string.

## Fix

Extract the per-extension check into a small helper `isMinikubeExtension` that wraps the `ToUnstructured` call with `recover()`. A malformed/unparseable extension is treated as non-minikube and the build proceeds normally. No behaviour change on the happy path.

## Reproduction

1. Update a cluster's extension to be string-valued
2. Attempt e.g. `devspace build`
3. Observe the panic

Confirmed on devspace v6.3.2 and v6.3.20.

Resolves #3218.

### DIFF
--- a/pkg/devspace/kubectl/util.go
+++ b/pkg/devspace/kubectl/util.go
@@ -272,6 +272,11 @@ func IsMinikubeKubernetes(kubeClient Client) bool {
 // panics on such values via reflection ("reflect.Set: value of type string is
 // not assignable to type map[string]interface {}") rather than returning an
 // error, so we recover() and treat unparseable extensions as non-minikube.
+//
+// noinline is required: if the compiler inlines this function into its caller,
+// the deferred recover() loses its stack frame and cannot catch the panic.
+//
+//go:noinline
 func isMinikubeExtension(extension runtime.Object) (result bool) {
 	defer func() {
 		if recover() != nil {
--- a/pkg/devspace/kubectl/util.go
+++ b/pkg/devspace/kubectl/util.go
@@ -255,18 +255,35 @@ func IsMinikubeKubernetes(kubeClient Client) bool {
 	if rawConfig, err := kubeClient.ClientConfig().RawConfig(); err == nil {
 		clusters := rawConfig.Clusters[rawConfig.Contexts[rawConfig.CurrentContext].Cluster]
 		for _, extension := range clusters.Extensions {
-			ext, err := runtime.DefaultUnstructuredConverter.ToUnstructured(extension)
-			if err == nil {
-				if provider, ok := ext["provider"].(string); ok {
-					if provider == minikubeProvider {
-						return true
-					}
-				}
+			if isMinikubeExtension(extension) {
+				return true
 			}
 		}
 	}
 
 	return false
+}
+
+// isMinikubeExtension safely checks whether a kubeconfig cluster extension
+// identifies the cluster as a minikube provider.
+//
+// Some tools (e.g. Teleport) write extension values as plain YAML strings
+// rather than structured objects. runtime.DefaultUnstructuredConverter.ToUnstructured
+// panics on such values via reflection ("reflect.Set: value of type string is
+// not assignable to type map[string]interface {}") rather than returning an
+// error, so we recover() and treat unparseable extensions as non-minikube.
+func isMinikubeExtension(extension runtime.Object) (result bool) {
+	defer func() {
+		if recover() != nil {
+			result = false
+		}
+	}()
+	ext, err := runtime.DefaultUnstructuredConverter.ToUnstructured(extension)
+	if err != nil {
+		return false
+	}
+	provider, ok := ext["provider"].(string)
+	return ok && provider == minikubeProvider
 }
 
 // GetKindContext returns the kind cluster name
--- a/pkg/devspace/kubectl/util.go
+++ b/pkg/devspace/kubectl/util.go
@@ -2,6 +2,7 @@ package kubectl
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


benmosher:
FWIW -- I am not a Go guy, so this is 100% Claude Code 😅 but this is a real problem I'm experiencing locally.

and I rebuilt the binary against this patch, and I can build again 👍 so the fix does empirically work.

pkg/devspace/kubectl/util.go
```diff
@@ -255,20 +255,42 @@ func IsMinikubeKubernetes(kubeClient Client) bool {
...
+			result = false
+		}
+	}()
+	ext, err := runtime.DefaultUnstructuredConverter.ToUnstructured(extension)
+	if err != nil {
+		return false
+	}
+	provider, ok := ext["provider"].(string)
+	return ok && provider == minikubeProvider
+}
```
danielbdias: Here we are basically doing a "kind-of" try/catch when the method panics, which might hide other errors happening here with k8s runtime lib ("k8s.io/apimachinery/pkg/runtime").

What we could do is to detect if `extension` is a plain string here and add into a format `runtime.DefaultUnstructuredConverter.ToUnstructured` might handle. What do you think?

benmosher: ahh smart. because I can better speak Python, something like `if isinstance(extension, Mapping):` test vs `try: extension["provider"]; except TypeError | KeyError | ...`

benmosher: @danielbdias what do you think of it now? switched to using an explicit JSON unmarshal.

benmosher:
#3269 appears to be attempting to resolve this as well. I haven't compared the implementations but would love to see this fixed somehow 🙏

stoposto:
We just ran into this issue as well and would love to see it fixed. @danielbdias can we please move this forward? Happy to help however I can 🙏

rostaklein:
Huh, when is this coming pleeease? 👀 We are struggling a bit.

PR #3219, opened 2026-04-21
Adds check for websocket source

zerbitx:
/kind bugfix

resolves #DSP-176

See: https://github.com/devspace-sh/devspace/security/advisories/GHSA-hqwm-7x7x-8379

The changes pertaining to the CVE are all to do with the websocket upgrader and logs, but ci was failing for a variety of reasons.  I kept it a several commits rather than forcing them all into one to make review easier.  The windows unit test failure is due to helm v4 already being on the runner, so it doesn't download v3 and the version string check fails.  I can add something to the workflow to remove the default installed helm so it will pass, but I figured we could also just leave that to a future PR.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -24,8 +24,8 @@ jobs:
     strategy:
       matrix:
         kubernetes_version:
-          - v1.27.3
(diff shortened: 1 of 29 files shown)


lizardruss (approved):
lgtm

PR #3231, opened 2026-04-23
Combined dependabot go

zerbitx:
/kind enhancement

resolves #


zerbitx:
## Claude review

  Critical (1)

  C1 — Nil dereference panic in termSizeQueueAdapter (exec.go:72)

  t.MonitorSize() returns nil when stdout isn't a real terminal. The old code assigned this directly to a
  remotecommand.TerminalSizeQueue interface, producing a nil interface (safe). The new code wraps it:
  &termSizeQueueAdapter{q: nil} — a non-nil interface holding a struct with a nil inner queue. When the
  executor calls sizeQueue.Next() → a.q.Next() → panic. Rare in production since t.Raw == true usually
  implies a real TTY, but triggerable in tests or with piped stdin.

  Fix: if q := t.MonitorSize(t.GetSize()); q != nil { sizeQueue = &termSizeQueueAdapter{q: q} }

  ---
  High (3)

  - H1 — overrideOnce (sync.Once) is never resettable, making the runtime error handler non-testable across
  test runs
  - H2 — No test for newRuntimeErrorHandler or OverrideRuntimeErrorHandler end-to-end; only the format
  helper is tested
  - H3 — gorilla/websocket is pinned to a pre-release pseudo-version (v1.5.4-0.20250319132907-e064f32e3674)
  — no CVE tracking, no stable release guarantees

  Medium (5)

  - M1 — CustomNavLink active detection silently changed from explicit startsWith to React Router v6's
  built-in matcher; end=false default undocumented for leaf routes
  - M2 — Dead isDevelopment comment in webpack dev config misleads future editors
  - M3 — Both webpack configs use the deprecated postcss-loader v3 plugins() API (will break on v5 upgrade)
  - M4 — terser compress.warnings: false is silently ignored in Terser v5
  - M5 — Post-wait pods.Items[0] access in inject.go is only safe because ExpectNoError aborts; implicit
  dependency

  Low (4)

  - L1 — No var _ remotecommand.TerminalSizeQueue = (*termSizeQueueAdapter)(nil) compile-time assertion
  - L2 — _ = upgradeRoundTripper.Close() discards error silently with no comment
  - L3 — Missing formatRuntimeError(nil, "") test case (switch default branch uncovered)
  - L4 — silenceDeprecations: ['legacy-js-api'] masks the Dart Sass migration rather than fixing it

  Full details saved to claude-review.md. C1 must be fixed before merge — the nil dereference is a real
  crash path.

zerbitx:
Lint is expected to fail due to the number of files in the commit, its finding old linting errors, despite the config for only new.

PR #3233, opened 2026-04-24
Dsp 173

zerbitx:
/kind bugfix
/kind enhancement
/kind feature
/kind documentation
/kind test

resolves #

### DIFF
--- a/cmd/flags/flags.go
+++ b/cmd/flags/flags.go
@@ -11,6 +11,7 @@ type GlobalFlags struct {
 	NoWarn                   bool
 	NoColors                 bool
 	Debug                    bool
+	ProtectUI                bool
 	DisableProfileActivation bool
 	SwitchContext            bool
 	InactivityTimeout        int
@@ -48,6 +49,7 @@ func SetGlobalFlags(flags *flag.FlagSet) *GlobalFlags {
 	flags.BoolVar(&globalFlags.NoWarn, "no-warn", false, "If true does not show any warning when deploying into a different namespace or kube-context than before")
 	flags.BoolVar(&globalFlags.NoColors, "no-colors", false, "Do not show color highlighting in log output. This avoids invisible output with different terminal background colors")
 	flags.BoolVar(&globalFlags.Debug, "debug", false, "Prints the stack trace if an error occurs")
+	flags.BoolVar(&globalFlags.ProtectUI, "protect-ui", false, "Enable UI protections such as auth checks for sensitive routes and redaction of sensitive config values")
 	flags.BoolVar(&globalFlags.Silent, "silent", false, "Run in silent mode and prevents any devspace log output except panics & fatals")
 
 	flags.StringSliceVarP(&globalFlags.Profiles, "profile", "p", []string{}, "The DevSpace profiles to apply. Multiple profiles are applied in the order they are specified")
--- a/cmd/run_pipeline.go
+++ b/cmd/run_pipeline.go
@@ -459,7 +459,7 @@ func runPipeline(ctx devspacecontext.Context, args []string, options *CommandOpt
 	})
 
 	// start ui & open
-	serv, err := dev.UI(ctx, options.UIPort, options.ShowUI, pipe)
+	serv, err := dev.UI(ctx, options.UIPort, options.ShowUI, options.ProtectUI, pipe)
 	if err != nil {
 		return err
 	}
--- a/cmd/ui.go
+++ b/cmd/ui.go
@@ -121,9 +121,9 @@ func (cmd *UICmd) RunUI(f factory.Factory) error {
 					continue
 				}
 
-				if serverVersion.DevSpace {
+				if serverVersion.DevSpace && serverVersion.Protected == cmd.ProtectUI {
 					cmd.log.Infof("Found running UI server at %s", domain)
-					_ = open.Start(domain)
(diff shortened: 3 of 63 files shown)


zerbitx:
Adversarial Review — PR #3233 (DSP-173 / UI protection)

  All tests pass, the build is clean, and the overall design (opt-in flag, 32-byte random token,
  SameSite=Strict cookie, constant-time comparison) is sound. The findings below are ordered by severity.

  ---
  MEDIUM — RawConfig returned unredacted when --protect-ui is active

  File: pkg/devspace/server/server.go:278-279

  retConfig.RawConfig = h.ctx.Config().Raw()
  retConfig.Config = h.ctx.Config().Config()

  Only Vars are redacted. RawConfig is the verbatim parsed devspace.yaml, which can contain hardcoded
  credentials (registry passwords, image pull secrets, build args) written outside the variable system. The
  flag description promises to "redact sensitive config values" — a user reading that would reasonably
  expect the raw config to be sanitised too, or at least understand what is and isn't covered. Either redact
   both, or tighten the flag description.

  ---
  MEDIUM — /api/logs-multiple referenced in UI, never registered on server

  File: ui/src/components/views/Logs/TerminalCache/TerminalCache.tsx:62

  withAuthQuery is applied to a /api/logs-multiple WebSocket URL:

  url: withAuthQuery(`${ApiWebsocketProtocol()}://${ApiHostname()}/api/logs-multiple?...`)

  This route is not registered anywhere in registerRoutes() — not in this PR, and not on main either. That
  makes the withAuthQuery call dead code. Either the backend route is missing from registerRoutes() (and
  should be added under handleAuthenticated), or the UI reference is stale and should be removed. As
  written, multi-log streaming silently falls back to a 404 when --protect-ui is enabled.

  ---
  LOW — Auth token persisted before server begins listening

  File: pkg/devspace/server/server.go:94-100

  NewServer writes the token to disk and then returns the *Server to the caller, who then calls
  ListenAndServe. If the server fails to start after token persistence (e.g., the port is stolen between
  findAvailablePort and listen), the stale token file remains. A subsequent devspace ui reads that file,
  constructs a URL with the stale token, and opens a browser to a dead address. Move persistAuthToken to
  happen only after the listener is successfully bound (or at least guard with removeAuthToken in the error
  path of ListenAndServe).

  ---
  LOW — Token file survives SIGKILL

  File: pkg/devspace/server/server.go:106-108

  server.Server.RegisterOnShutdown(func() {
      _ = removeAuthToken(server.Server.Addr)
  })

  RegisterOnShutdown callbacks only fire during graceful http.Server.Shutdown(). A process killed with
  SIGKILL (OOM killer, container restart, kill -9) leaves the token file on disk. The next devspace ui reads
   it, gets a URL that points at nothing, and opens a dead browser tab. Low security risk (no running server
   to attack), but the UX is confusing. A defer removeAuthToken(...) installed with signal.NotifyContext or
  an os.Signal handler would give broader coverage.

  ---
  LOW — Silent skip when protection level mismatches

  File: cmd/ui.go:124

  if serverVersion.DevSpace && serverVersion.Protected == cmd.ProtectUI {

  When devspace dev --protect-ui is running and the user invokes devspace ui (without --protect-ui), the
  discovery loop silently bypasses the existing server and starts a new one on the next available port.
      _ = removeAuthToken(server.Server.Addr)
  })

  RegisterOnShutdown callbacks only fire during graceful http.Server.Shutdown(). A process killed with
  SIGKILL (OOM killer, container restart, kill -9) leaves the token file on disk. The next devspace ui reads
   it, gets a URL that points at nothing, and opens a dead browser tab. Low security risk (no running server
   to attack), but the UX is confusing. A defer removeAuthToken(...) installed with signal.NotifyContext or
  an os.Signal handler would give broader coverage.

  ---
  LOW — Silent skip when protection level mismatches

  File: cmd/ui.go:124

  if serverVersion.DevSpace && serverVersion.Protected == cmd.ProtectUI {

  When devspace dev --protect-ui is running and the user invokes devspace ui (without --protect-ui), the
  discovery loop silently bypasses the existing server and starts a new one on the next available port.
  There is no log message explaining why. Users will see two servers running with no explanation. The fix is
   a one-line warning: "Found a running UI server but its protection level does not match; starting a new
  server".

  ---
  LOW — isSensitiveVariableName pluralisation stripping is fragile

  File: pkg/devspace/server/server.go:339-342

  upper := tokenUpper
  if len(upper) > 1 {
      upper = strings.TrimSuffix(upper, "S")
  }

  This handles CREDENTIALS→CREDENTIAL correctly but also strips the trailing S from words like
  "PROCESS"→"PROCES" or "STATUS" depending on the identifier. It works correctly for the tested cases but
  the approach is brittle. Consider matching the keywords as prefixes or suffixes against the token
  (strings.Contains(upper, kw)) uniformly instead of mutating upper before the switch.

  ---
  INFO — PR template not filled in

  The PR body still has placeholder text: "resolves #", "Fixed an issue where DevSpace ...". No Linear issue
   is linked, and the release note copy is a template stub. This blocks changelog generation and makes it
  harder to trace the work back to an issue.

  ---
  INFO — Docs only show devspace dev and devspace ui

  File: docs/pages/getting-started/development.mdx

  --protect-ui is a global flag that also applies to devspace run-pipeline (and any other pipeline command
  that starts the UI). The docs only show the two explicit examples. A note like "The --protect-ui flag is
  available on all commands that start the UI" avoids user confusion when they try devspace run-pipeline
  --protect-ui and find no docs for it.

PR #3256, opened 2026-05-01
Parallelize import download/read/unmarshal work

zerbitx:
Replace global dependency download locking with per-cache locking and in-flight dedupe Add regression coverage for concurrent imports, merge order, disabled imports, and shared git subpaths Make URL dependency downloads context-aware and document download synchronization behavior

/kind enhancement

resolves #DSP-152

### DIFF
--- a/e2e/tests/imports/imports.go
+++ b/e2e/tests/imports/imports.go
@@ -139,6 +139,30 @@ var _ = DevSpaceDescribe("imports", func() {
 		framework.ExpectError(err)
 	})
 
+	ginkgo.It("should import multiple git subpaths from the same repository", func() {
+		tempDir, err := framework.CopyToTempDir("tests/imports/testdata/git-subpaths")
+		framework.ExpectNoError(err)
+		defer framework.CleanupTempDir(initialDir, tempDir)
+
+		configBuffer := &bytes.Buffer{}
+		printCmd := &cmd.PrintCmd{
+			GlobalFlags: &flags.GlobalFlags{},
+			Out:         configBuffer,
+			SkipInfo:    true,
+		}
+
+		err = printCmd.Run(f)
+		framework.ExpectNoError(err)
+
+		latestConfig := &latest.Config{}
+		err = yaml.Unmarshal(configBuffer.Bytes(), latestConfig)
+		framework.ExpectNoError(err)
+
+		framework.ExpectEqual(latestConfig.Pipelines["pipeline-a"].Run, "echo \"pipeline-a\"")
+		framework.ExpectEqual(latestConfig.Pipelines["pipeline-b"].Run, "echo \"pipeline-b\"")
+		framework.ExpectEqual(latestConfig.Pipelines["pipeline-c"].Run, "echo \"pipeline-c\"")
+	})
+
 	ginkgo.It("should import correctly with localRegistry", func() {
 		tempDir, err := framework.CopyToTempDir("tests/imports/testdata/localregistry")
 		framework.ExpectNoError(err)
--- a/e2e/tests/imports/testdata/git-subpaths/devspace.yaml
+++ b/e2e/tests/imports/testdata/git-subpaths/devspace.yaml
@@ -3,11 +3,11 @@ name: git-subpath-imports
 
 imports:
   - git: https://github.com/loft-sh/e2e-test-dependency.git
-    branch: main
+    revision: 1ec44dcffc83f64ea7e8c169751e56d233cd4b20
     subPath: subA
   - git: https://github.com/loft-sh/e2e-test-dependency.git
-    branch: main
+    revision: 1ec44dcffc83f64ea7e8c169751e56d233cd4b20
     subPath: subB
   - git: https://github.com/loft-sh/e2e-test-dependency.git
-    branch: main
+    revision: 1ec44dcffc83f64ea7e8c169751e56d233cd4b20
     subPath: subC
--- a/e2e/tests/imports/testdata/git-subpaths/devspace.yaml
+++ b/e2e/tests/imports/testdata/git-subpaths/devspace.yaml
@@ -1,0 +1,13 @@
+version: v2beta1
+name: git-subpath-imports
+
+imports:
+  - git: https://github.com/loft-sh/e2e-test-dependency.git
+    branch: main
+    subPath: subA
+  - git: https://github.com/loft-sh/e2e-test-dependency.git
+    branch: main
+    subPath: subB
+  - git: https://github.com/loft-sh/e2e-test-dependency.git
+    branch: main
+    subPath: subC
--- a/pkg/devspace/config/loader/imports.go
+++ b/pkg/devspace/config/loader/imports.go
@@ -8,11 +8,13 @@ import (
 
 	"github.com/loft-sh/devspace/pkg/devspace/config/loader/variable"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions"
+	"github.com/loft-sh/devspace/pkg/devspace/config/versions/latest"
 	"github.com/loft-sh/devspace/pkg/devspace/config/versions/util"
 	dependencyutil "github.com/loft-sh/devspace/pkg/devspace/dependency/util"
 	"github.com/loft-sh/devspace/pkg/util/log"
 	"github.com/loft-sh/devspace/pkg/util/yamlutil"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
 var ImportSections = []string{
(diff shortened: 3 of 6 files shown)


zerbitx:
From Claude:

Critical Issues

  1. Incorrect comment in downloadCallKey — wrong reasoning, correct behavior

  pkg/devspace/dependency/util/util.go (line ~334):
  // SubPath is encoded in id (via GetDependencyID) and does not need to be
  // listed separately.

  This is false. GetDependencyID for git sources only encodes Git + Branch/Tag/Revision — SubPath is not included. Two imports with the same git URL but different subpaths (subA, subB, subC) produce the same ID and the same downloadCallKey.

  The code behavior is actually correct — both imports share one clone, then getDependencyConfigPath applies SubPath afterward to resolve each import's config file within the shared repo. But the comment gives the wrong explanation. This will confuse anyone reasoning about deduplication behavior later. The comment should be:

  ▎ SubPath navigates within an already-cloned repo; all imports sharing the same git URL write to the same cache directory, so SubPath need not distinguish download calls.

  ---
  High Issues

  2. runDownloadOnce is not panic-safe — goroutine leak under panic

  util.go lines ~285–300:
  call.err = download()
  close(call.done)         // <— never reached if download() panics
  downloadCallsMutex.Lock()
  delete(downloadCalls, key)
  downloadCallsMutex.Unlock()

  If download() panics, call.done is never closed. Any goroutine blocked on <-call.done will hang forever. Fix:

  defer func() {
      close(call.done)
      downloadCallsMutex.Lock()
      delete(downloadCalls, key)
      downloadCallsMutex.Unlock()
  }()
  call.err = download()

  3. Dead code guard in imports.go:87–88

  } else if resolvedImport.Data == nil {
      return nil, errors.Errorf("resolved import is missing data")
  }

  For any non-disabled import, loadImports either sets Data to a non-nil map or returns an error (which prevents the merge loop from running). This branch is unreachable. It should be removed — dead error paths confuse readers about invariants.

  ---
  Medium Issues

  4. Concurrent download test is timing-sensitive and could be flaky

  imports_test.go:TestResolveImportsDownloadsSiblingImportsConcurrently — the test blocks all 3 HTTP handlers until all 3 are simultaneously active, using time.After(2 * time.Second) as a fallback. On a heavily loaded CI machine, goroutines may not be scheduled within 2 seconds, causing all 3 to unblock sequentially. The final
  assert.Assert(t, maxActive == 3) would then fail. No mechanism retries or makes the timing deterministic.

  Consider using a semaphore-based synchronization barrier with explicit "all N requests in flight" acknowledgment rather than wall-clock timeout.

  5. No test for runDownloadOnce error propagation to concurrent waiters

  util_test.go:TestRunDownloadOnceCleansUpCall only verifies cleanup on success. There is no test verifying that when a download returns an error, concurrent waiters also receive that error (not a stale nil or wrong value). The call.err = download() path is untested under concurrent waiter conditions.

  6. maxConcurrentImportDownloads = 8 is unexplained

  imports.go:33:
  const maxConcurrentImportDownloads = 8

  No comment explaining this choice. Is it benchmarked? Based on typical network concurrency? A typo for another value? Without explanation, the next engineer will have no idea whether to increase or decrease it for performance tuning.

  7. Trailing whitespace in util.go


  4. Concurrent download test is timing-sensitive and could be flaky

  imports_test.go:TestResolveImportsDownloadsSiblingImportsConcurrently — the test blocks all 3 HTTP handlers until all 3 are simultaneously active, using time.After(2 * time.Second) as a fallback. On a heavily loaded CI machine, goroutines may not be scheduled within 2 seconds, causing all 3 to unblock sequentially. The final
  assert.Assert(t, maxActive == 3) would then fail. No mechanism retries or makes the timing deterministic.

  Consider using a semaphore-based synchronization barrier with explicit "all N requests in flight" acknowledgment rather than wall-clock timeout.

                                                                                                                                                                                                      5. No test for runDownloadOnce error propagation to concurrent waiters

  util_test.go:TestRunDownloadOnceCleansUpCall only verifies cleanup on success. There is no test verifying that when a download returns an error, concurrent waiters also receive that error (not a stale nil or wrong value). The call.err = download() path is untested under concurrent waiter conditions.
  6. maxConcurrentImportDownloads = 8 is unexplained

                                                                                                                                                                                                                                                                                                    imports.go:33:                                                                                                                                                                                                                                                                                                                                        const maxConcurrentImportDownloads = 8
                                                                                                                                                                                                                                                                                                                No comment explaining this choice. Is it benchmarked? Based on typical network concurrency? A typo for another value? Without explanation, the next engineer will have no idea whether to increase or decrease it for performance tuning.

  7. Trailing whitespace in util.go

                                                                                                                                                                                                                                                                                                                     Two instances in downloadURLDependency:
  log.Warnf("Error closing body: %v", err)\t   // trailing tab                                                                                                                                                                                                                                                                                          }()·                                           // trailing space

                                                                                                                                                                                                                                                                                      8. e2e test references an unpinned external repository

                                                                                                                                                                                                                                                                                                testdata/git-subpaths/devspace.yaml:
  - git: https://github.com/loft-sh/e2e-test-dependency.git                                                                                                                                                                                                                                                                                               branch: main
    subPath: subA

                                                                                                                                                                                                                                                                                                                                       No commit SHA pinned. If loft-sh/e2e-test-dependency:main is modified, the e2e test silently changes semantics or breaks. Should pin a specific commit for reproducibility, or document this is intentional.

  ---                                                                                                                                                                                                                                                                                                                                                   Low Issues

  9. reloadVariables is only called in the else branch, not after imports-with-subimports
  When an import has its own nested imports, ResolveImports is called recursively (which calls reloadVariables at its start). When an import has no nested imports, reloadVariables is called explicitly. This logic is equivalent to the original but subtle — if the branch structure changes in the future, it's easy to miss the call. A comment
  explaining why the if branch doesn't need an explicit reloadVariables call would help.
  10. loadImports uses index-loop capture fix but comment not present

  imports.go:153–154:                                                                                                                                                                                                                                                                                                                                   for index := range imports {
      importConfig := imports[index]
  The pre-capture of importConfig (rather than referencing imports[index] inside the closure) is the correct Go fix for loop variable capture, but this is not the current Go 1.22+ behavior where range variables are per-iteration. A brief comment noting this is intentional would avoid a future "cleanup" that reintroduces the bug.
  ---                                                                                                                                                                                                                                                                                                                                                   Positive Findings

                                                                                                                                                                                                                                                                                                                                     - withDownloadLock's defer ordering is subtle but correct and well-commented: mutex.Unlock() runs before releaseDownloadLock (LIFO), ensuring the lock is released before the ref is decremented, preventing premature deletion of a lock still in use by a waiter.                                                                                   - Context propagation to HTTP requests (http.NewRequestWithContext) is a genuine improvement over the old http.Get.
  - The errgroup usage (SetLimit, index pre-capture) is clean and idiomatic.                                                                                                                                                                                                                                                                            - The ordered-merge-semantics test (TestResolveImportsKeepsOrderedMergeSemantics) is a good regression guard — it explicitly documents first-import-wins for key conflicts and append-semantics for slices.
  - The TestResolveImportsSkipsDisabledImports test uses a nonexistent file as the disabled import path, which correctly verifies that disabled imports are never downloaded.

  ---                                                                                                                                                                                                                                                                                                                                                   Summary

  The parallelization approach is architecturally sound and the core locking design is correct. Before merge: fix the panic-safety gap in runDownloadOnce, remove the dead guard in the merge loop, correct the misleading comment in downloadCallKey, and add a test for error propagation to concurrent waiters. The timing-sensitive e2e             concurrency test and the unpinned external git dependency are worth addressing but lower priority.

zerbitx:
Follow up review comments:

  New code examined — no new issues found
                                                                                                                                                                                                                                                                                                               - The panic recovery path (call.err = fmt.Errorf("download panicked: %v", panicValue) → cleanup() → panic(panicValue)) is correctly ordered: call.err is set before close(call.done) inside cleanup, satisfying the happens-before required for waiters to safely read the error.
  - cleanup is invoked exactly once per call in the deferred func (either the panic branch or the normal branch) — no double-close risk.                                                                                                                                                                                                                - TestRunDownloadOnceReturnsErrorToConcurrentWaiter correctly relies on the Go channel happens-before guarantee: close(call.done) is sequenced after call.err = expectedErr, so the waiter's read is safe.
  - The new concurrent test is now deterministic: <-allInFlight guarantees all 3 were simultaneously in-flight before the assertion, so maxActive == 3 is not just probabilistic.
  ---
  Verdict

                                                                                                                                                                                                                                                                                                                                               The implementation is sound. All substantive issues from the previous review are addressed. The only remaining item before marking ready is a one-line comment on maxConcurrentImportDownloads. Good to go after that.
                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                               

💡 I'm skipping the comment as the variable name is clear enough.

PR #3295, opened 2026-08-24
fix: close HTTP response bodies in helper requests

wangjingshuiku:

/kind bugfix

Closes HTTP response bodies in the URL readiness polling loop and after downloading the restart helper. This prevents connections and file descriptors from being retained across retries. No associated issue.

Fixed an issue where repeated URL readiness checks and restart-helper downloads could retain HTTP resources.

`go test ./pkg/devspace/build/builder/restart` passes. The changed `cmd` package was formatted with `gofmt`, and `git diff --check` passes.


wangjingshuiku:
@FabianKramm @LukasGentele Could you please take a look when you have a chance? Thanks!

